PR #3, opened 2020-05-04
Calculate start date of new price for each subscription

kelvin-chappell:
At the core of this change, there's a slightly stripped down billing date calculation ripped from support-service-lambdas [here](https://github.com/guardian/price-migration-engine/blob/b68411d8dfcfc9a77b2a1f62b1dcee47946ed0bb/src/main/scala/pricemigrationengine/model/BillingDate.scala).  Apart from that the fields needed to calculate the billing date are added to the [subscription model](https://github.com/guardian/price-migration-engine/blob/b68411d8dfcfc9a77b2a1f62b1dcee47946ed0bb/src/main/scala/pricemigrationengine/model/ZuoraSubscription.scala).  And I've added the first [test case](https://github.com/guardian/price-migration-engine/blob/118cd48b646324720d9f57d720e37587980e1f6d/src/test/resources/MonthlyDiscounted.json).


src/main/scala/pricemigrationengine/model/BillingDate.scala
```diff
@@ -0,0 +1,565 @@
+package pricemigrationengine.model
```
frj, line 1: As you say this is quite a chunky bit of code to copy over from support-service-lambdas.

I dont think we have 100% confidence in this code being able to predict the next bill date for every case as demonstrated by the alerts we get from  zuora-creditor.

Its probably close enough for this use case, although we could try using the zuora billing preview endpoint. In retrospect thats probably the route I would have gone down in support service lambdas.

If we bring this code across we should definitely bring over the test cases as well, ie a test for each of the different 'types' of subs ie:

https://github.com/guardian/support-service-lambdas/tree/master/lib/zuora-core/src/test/resources

It should just be a case of bringing over the tests from support-service-lambdas and adapting slightly.

kelvin-chappell: I'll bring the test cases in.  I'm going to add all the variants of the voucher cases as well.  So I'll do that in this PR.
The complexity of the billing date code is a bit alarming.  I'll add a card to investigate the billing preview too.

kelvin-chappell:
I think I'll just bring in one more test case for a sub without a discount and then move on to trying out billing preview.

PR #9, opened 2020-05-07
Implement dynamodb for cohort table

frj:
This PR contains the code to read values from the CohortTable from DynamoDB.

The code uses the aws java api client to call the DynamoDB query endpoint. The query endpoint returns a batch with a getLastEvaluatedKey which is used in the query to get the next batch.

The batching happens transparently inside a ZStream.

This is work in progress but the changes were getting large so it made sense to create the PR now.

TODO:
- Implement the write operation
- Add environment specific creation of CohortTable name
- Implement tests


src/main/scala/pricemigrationengine/services/CohortTable.scala
```diff
@@ -1,19 +1,69 @@
...
 
-  def fetch(filter: CohortTableFilter, batchSize: Int): ZIO[CohortTable, CohortFetchFailure, Set[CohortItem]] =
+  def fetch(filter: CohortTableFilter, batchSize: Int): ZIO[CohortTable, CohortFetchFailure, ZStream[Any, CohortFetchFailure, CohortItem]] =
     ZIO.accessM(_.get.fetch(filter, batchSize))
 
   def update(result: EstimationResult): ZIO[CohortTable, CohortUpdateFailure, Unit] =
     ZIO.accessM(_.get.update(result))
+
+  val impl: ZLayer[DynamoDBZIO, Nothing, UserRepo] =
+    ZLayer.fromFunction { dynamoDBZStream: DynamoDBZIO =>
```
frj, line 32: Im interested in peoples opinions of this pattern. 

It seems if you use the layer 'Acccessor' methods eg [here](https://github.com/guardian/price-migration-engine/blob/683dd57b47802124dbe50d69dd41d6b38e9281fa/src/main/scala/pricemigrationengine/services/Logging.scala#L13) in the service methods eg in this case 'fetch' it forces you to include the Logging service as an input for the IO returned by the service method.

Eg the fetch interface becomes:
```
def fetch(... ): ZIO[DynamoDBZIO, CohortFetchFailure, ZStream[Any, CohortFetchFailure, CohortItem]]
```

Which is exposing the implementation in the interface and there would be a different interface for different implementations.

Its weird as it seems to make the accessors only useful at the top level.

Am I missing something, is there a better way of doing this.

src/main/scala/pricemigrationengine/services/package.scala
```diff
@@ -5,5 +5,6 @@ import zio.Has
 package object services {
   type CohortTable = Has[CohortTable.Service]
   type Zuora = Has[Zuora.Service]
+  type UserRepo = Has[CohortTable.Service]
```
kelvin-chappell, line 8: `UserRepo`?

dynamoDb/README.md
```diff
@@ -0,0 +1,51 @@
...
+#Running DynamoDB Locally
+
+Install and run dynamoDB on you local machine using the following instructions:
+- [https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html)
+
+
+Create the CohortTable using the following aws cli command:
+```$bash
+aws dynamodb create-table \
+    --region us-east-1 \
```
arausuy, line 11: not eu-west-1?

src/main/scala/pricemigrationengine/services/CohortTable.scala
kelvin-chappell: I think the point is to distinguish between the environment required to build a service and the environment required to use the service.  So, for example, a live implementation would generally need a configuration service to build it but the interface wouldn't need to know about configuration as it's not needed to make use of the service.
So Logging should probably be part of the interface spec unless the service only logs while it's being built.

src/main/scala/pricemigrationengine/dynamodb/DynamoDBZIO.scala
```diff
@@ -0,0 +1,62 @@
...
+
+object DynamoDBZIO {
+  trait Service {
+    def query[A](
+      query: QueryRequest,
+      deserializer: java.util.Map[String, AttributeValue] => IO[DynamoDBZIOError, A]
+    ): ZStream[Any, DynamoDBZIOError, A]
+  }
+
+  val impl:  ZLayer[DynamoDBClient with Logging, Nothing, DynamoDBZIO] =
```
kelvin-chappell, line 24: Perhaps keep the implementation separate from the interface so that it's clearer to distinguish.  I know that in the ZIO docs they seem to put the main implementation with the spec but might be clearer apart.

src/main/scala/pricemigrationengine/dynamodb/DynamoDBZIO.scala
```diff
@@ -0,0 +1,62 @@
...
+    ZLayer.fromFunction { dependencies: DynamoDBClient with Logging =>
```
kelvin-chappell, line 25: I think the `Logging` is a dependency of the service rather than its creation.

src/main/scala/pricemigrationengine/dynamodb/DynamoDBClient.scala
```diff
@@ -0,0 +1,28 @@
...
+            )
+            .mapError(ex => s"Failed to create the dynamoDb client: $ex")
+        ) { dynamoDB: AmazonDynamoDB =>
+          ZIO
+            .effect(dynamoDB.shutdown)
+            .catchAll(ex => logging.get.error(s"Failed to close dynamo db connection: $ex"))
+        }
+      }
+    )
+}
```
kelvin-chappell, line 28: Could this file go in the services package? Or in `services/dynamodb`?

src/main/scala/pricemigrationengine/services/CohortTable.scala
```diff
@@ -1,19 +1,69 @@
...
   }
 
-  def fetch(filter: CohortTableFilter, batchSize: Int): ZIO[CohortTable, CohortFetchFailure, Set[CohortItem]] =
+  def fetch(filter: CohortTableFilter, batchSize: Int): ZIO[CohortTable, CohortFetchFailure, ZStream[Any, CohortFetchFailure, CohortItem]] =
     ZIO.accessM(_.get.fetch(filter, batchSize))
 
   def update(result: EstimationResult): ZIO[CohortTable, CohortUpdateFailure, Unit] =
     ZIO.accessM(_.get.update(result))
+
+  val impl: ZLayer[DynamoDBZIO, Nothing, UserRepo] =
```
kelvin-chappell, line 31: Perhaps keep the implementation separate from the interface so that it's clearer to distinguish?

src/main/scala/pricemigrationengine/dynamodb/DynamoDBZIO.scala
```diff
@@ -0,0 +1,62 @@
...
+        private def sendQueryRequest[A](queryRequest: QueryRequest): ZIO[Any, DynamoDBZIOError, QueryResult] = {
+          ZIO.effect {
+              logging.info(s"Starting query: $queryRequest")
+              amazonDynamoDB.query(queryRequest)
+            }
+            .mapError(ex => DynamoDBZIOError(s"Failed to execute query $queryRequest : $ex"))
+        }
+      }
+    }
+}
```
kelvin-chappell, line 62: Could this file go in the services package? Or in `services/dynamodb`?

src/main/scala/pricemigrationengine/dynamodb/package.scala
```diff
@@ -0,0 +1,9 @@
+package pricemigrationengine
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import zio.Has
+
+package object dynamodb {
+  type DynamoDBClient = Has[AmazonDynamoDB]
+  type DynamoDBZIO = Has[DynamoDBZIO.Service]
+}
```
kelvin-chappell, line 9: Could this file go in the services package? Or in `services/dynamodb`?

src/main/scala/pricemigrationengine/dynamodb/DynamoDBClient.scala
```diff
@@ -0,0 +1,28 @@
...
+object DynamoDBClient {
+  val dynamoDB: ZLayer[Logging, String, DynamoDBClient] =
+    ZLayer.fromManaged(
+      ZManaged.fromFunctionM { logging: Logging =>
+        ZManaged.make(
+          ZIO
+            .effect(
+              AmazonDynamoDBClient
+                .builder()
+//                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration("http://localhost:8000", "us-east-1"))
```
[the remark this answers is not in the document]
frj: Yes definitely makes sense. We are going to have to have some configuration ie 'Stage' so we use the correct dynamo table. Its in the todos, we can do it then?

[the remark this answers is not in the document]
kelvin-chappell: 👍

src/main/scala/pricemigrationengine/services/CohortTable.scala
frj: Hum... For example the ConsoleLogging service is dependent on the console which it does nothing with at creation time, only using it when you 'use' a logging method.

You wouldnt add it to the interface as you would need to supply a console dependency to the Lambda logging.

And looking at the implementation you have done the same thing as I did here, ie you dont use the Console 'accessor' methods as that seems to force you to add the dependencies to the interface.

Thinking about you could use the accessors and then provide the dependencies eg:
```scala
object ConsoleLogging {
  val impl: ZLayer[Console, Nothing, Logging] = ZLayer.fromFunction(
    consoleService =>
      new Logging.Service {
        def info(s: String): UIO[Unit] = console.putStrLn(s"INFO: $s").provide(consoleService)
        def error(s: String): UIO[Unit] = console.putStrLn(s"ERROR: $s").provide(consoleService)
    }
  )
}
``` 

But that maybe seems a bit less readable, not sure

kelvin-chappell: Yes I was thinking about it again, and from a method that has the service as part of its environment it might become a bit unwieldy if you have to include the dependencies of the services you're calling as well as the services you're making use of directly.  So you probably wouldn't expect the return type of a method to be `ZIO[... with Logging, ..., ...]` unless you were actually using `Logging` in your method.

So this is fine as it is. 👍

src/main/scala/pricemigrationengine/services/package.scala
frj: Opps missed that 👍

src/main/scala/pricemigrationengine/dynamodb/DynamoDBZIO.scala
kelvin-chappell: Ignore.

PR #10, opened 2020-05-11
Implementation for update item in CohortTable

frj:
Implements the updating of CohortTable entries with results of the price and start date estimation process.

- This uses the update dynamodb operation with will update the specified fields leaving any other fields unchanged
- Refactored the case class -> dynamo dbclient representation serialisation to use implicit serialisers


src/main/scala/pricemigrationengine/services/CohortTable.scala
```diff
@@ -28,31 +32,60 @@ object CohortTable {
...
+      getStringFromResults(cohortItem, "subscriptionNumber")
+        .map(CohortItem.apply)
+
+  private implicit val estimationResultSerialiser: DynamoDBUpdateSerialiser[EstimationResult] =
+    estimationResult =>
+      Map(
+        stringFieldUpdate("processingStage", EstimationComplete.value),
+        dateFieldUpdate("expectedStartDate", estimationResult.expectedStartDate),
+        doubleFieldUpdate("estimatedNewPrice", estimationResult.estimatedNewPrice)
+      ).asJava
```
kelvin-chappell, line 46: There are a couple of new fields now as well in the `estimationResult`: `currency` and `oldPrice`.

src/main/scala/pricemigrationengine/services/CohortTable.scala
```diff
@@ -28,31 +32,60 @@ object CohortTable {
...
-                .withLimit(batchSize),
-              { result =>
-                getStringFromResults(result, "subscriptionNumber").map(CohortItem.apply)
-              }
-            ).mapError(error => CohortFetchFailure(error.toString))
+            dynamoDBZIO
+              .get
+              .query(
+                new QueryRequest()
+                  .withTableName("PriceMigrationEngineDev")
```
kelvin-chappell, line 73: Might be easier if this name is configurable, as will probably have a different table for each cohort.

frj: Ill create a separate PR with the changes to add the configuration.

kelvin-chappell: 👍

PR #13, opened 2020-05-12
Implement Zuora service

kelvin-chappell:
The implementation is [here](https://github.com/guardian/price-migration-engine/blob/c0e2f2ee292044c77f6c8c1f6e85f0411c083143/src/main/scala/pricemigrationengine/services/ZuoraLive.scala), and it's wired into the environment [here](https://github.com/guardian/price-migration-engine/blob/c0e2f2ee292044c77f6c8c1f6e85f0411c083143/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala#L54).

I've used the [scalaj http library](https://github.com/guardian/price-migration-engine/blob/c0e2f2ee292044c77f6c8c1f6e85f0411c083143/project/Dependencies.scala#L11) as it's very lightweight and has no dependencies.  There are some advantages in the testability of the sttp library because it can be mocked more easily but I think everything that's worth testing could be extracted and made testable in isolation anyway.


src/main/scala/pricemigrationengine/services/ZuoraLive.scala
```diff
@@ -0,0 +1,85 @@
...
+                Seq(
+                  "grant_type" -> "client_credentials",
+                  "client_id" -> config.zuora.clientId,
+                  "client_secret" -> config.zuora.clientSecret
+                )
+              )
+          ).map(_.access_token)
+
+        private def get[A: Reader](path: String): ZIO[Any, ZuoraFetchFailure, A] =
+          fetchedAccessToken flatMap { accessToken =>
```
frj, line 39: maybe use a for comprehension

src/main/scala/pricemigrationengine/services/ZuoraLive.scala
```diff
@@ -0,0 +1,85 @@
...
+
+        private val readTimeout = HttpOptions.readTimeout(Duration(30, SECONDS).toMillis.toInt)
+
+        private case class AccessToken(access_token: String)
+        private implicit val rwAccessToken: ReadWriter[AccessToken] = macroRW
+
+        private case class InvoicePreviewRequest(accountId: String, targetDate: LocalDate)
+        private implicit val rwInvoicePreviewRequest: ReadWriter[InvoicePreviewRequest] = macroRW
+
+        private lazy val fetchedAccessToken: ZIO[Any, ZuoraFetchFailure, String] =
```
frj, line 26: I dont think this lazy val will have the expected effect ie the fetchAccessToken effect will get executed every time its referenced.

You could fetch the access token in the constructor.

There is an IO.memoize function, Im not 100% sure how to uses it, ie it gives you an nested IO which is a bit weird.

kelvin-chappell: That's a good point.  I can't make head or tail of how `memoize` is supposed to work; there's also a `once` method which is similarly opaque.  I think I'll just create it in the conventional way, as you suggest.

PR #14, opened 2020-05-12
Dynamo db config

frj:
This PR adds the additional configuration properties that need to be set as environment variables to run the EstimationHandler:
- Stage - Stage ie DEV/CODE/PROD, this is used to generate the correct cohort table for each environment
- dynamodb.serviceEndpoint - Optional parameter to override the dynamodb endpoint useful for using a local dynamodb
- dynamodb.signingRegion - region associated with the overridden service endpoint

### DIFF
--- a/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
+++ b/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
@@ -5,7 +5,7 @@ import java.time.LocalDate
 import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
 import pricemigrationengine.model.CohortTableFilter.ReadyForEstimation
 import pricemigrationengine.model._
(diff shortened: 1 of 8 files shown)


src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -5,7 +5,7 @@ import java.time.LocalDate
 import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
 import pricemigrationengine.model.CohortTableFilter.ReadyForEstimation
 import pricemigrationengine.model._
-import pricemigrationengine.services._
+import pricemigrationengine.services.{EnvConfiguration, _}
```
kelvin-chappell, line 8:
```suggestion
import pricemigrationengine.services._
```

PR #19, opened 2020-05-14
Add riffraff configuration for lambda

frj:
This PR adds the cloudformation and riffraff configuration to allow the deployment of the price-migration-engine-lambda.

Details include:
- Config is stored in the secret service and injected into the app using environment variables and and dynamic references in the cloudformation
- Changed the format of environment variable keys for compatibility with cloudformation


build.sbt
```diff
@@ -29,5 +30,12 @@ lazy val lambda = (project in file("lambda"))
       http,
       munit % Test
     ),
-    testFrameworks += new TestFramework("munit.Framework")
+    testFrameworks += new TestFramework("munit.Framework"),
+    description := "Lambda jar for the Price Migration Engine",
+    assemblyJarName := s"price-migration-engine-lambda.jar",
```
kelvin-chappell, line 35: Could reuse `name` or remove `s`

lambda/cfn.yaml
```diff
@@ -0,0 +1,75 @@
...
+        - PolicyName: DynamoDBAccessPolicy
+          PolicyDocument:
+            Statement:
+              - Effect: Allow
+                Action:
+                  - dynamodb:Query
+                  - dynamodb:UpdateItem
+                Resource:
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/PriceMigrationEngine${Stage}"
+                  - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/PriceMigrationEngine${Stage}/*"
```
kelvin-chappell, line 47: What does this line refer to?

lambda/cfn.yaml
```diff
@@ -0,0 +1,75 @@
...
+      FunctionName:
+        !Sub price-migration-engine-lambda-${Stage}
+      Code:
+        S3Bucket: membership-dist
+        S3Key: !Sub membership/${Stage}/price-migration-engine-lambda/price-migration-engine-lambda.jar
+      Handler: pricemigrationengine.handlers.EstimationHandler::handleRequest
+      Environment:
+        Variables:
+          Stage: !Ref Stage
+          zuoraApiHost: !Sub '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:zuoraApiHost}}'
```
kelvin-chappell, line 62: This one is just stored as a String

lambda/cfn.yaml
frj: It refers to the indexes which are required for the queries we do.

lambda/cfn.yaml
frj: Im using the secret manager instead of the parameter store 'secure-string' as cloud formation doesnt support using secure strings from the parameter store in environment variables.

I could put the non-secret config in the parameter store or have it in a Mapping in the cloudformation template its self, but its probably easier just to have it in one place?

kelvin-chappell: No, this makes sense.  I'm not familiar with the secrets manager.  Strange that AWS has two very similar storage mechanisms.

PR #21, opened 2020-05-15
Improve secret update process

frj:
This change improves the process for updating values stored in the secretsmanager.

Cloudformation will not update a resource if there are no changes to the resource in the cloudformation template.

Updating the secrets in the [secretsmanager](https://eu-west-1.console.aws.amazon.com/secretsmanager/home?region=eu-west-1#/secret?name=price-migration-engine-lambda-DEV) would not updated the values in the environment variables of the lambda, even after a redeploy of the app via riffraff.

This change adds the secrets version number to the dynamic references used to access the secrets which when changed force cloudformation  to update the resource with the values from the new version of the secrets.
 
For more info see:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html

### DIFF
--- a/lambda/README.md
+++ b/lambda/README.md
@@ -10,20 +10,34 @@ The configuration for this application is stored in the [aws secret manager](htt
 
 The configuration can be updated using the aws console as follows:
 
-- Log into the aws console using the 'membership' profile via [janus](https://janus.gutools.co.uk/)
-- Navigate to Services > Security, Identity, & Compliance > Secrets Manager
-- Search for price-migration-engine-<STAGE> and click on the result
-- Click 'Retrieve Secret Value' button
-- Click 'Edit' 
-- Add/Edit the key value pairs
-- Click 'Save'
-
-Cloudformation seems reluctant to reapply the updated configuration to the lambda environment variables. This 
-can be done by deleting the lambda stack and redeploying it with riffraff.
-
-You can delete the stack as follows:
-- Log into the aws console using the 'membership' profile via [janus](https://janus.gutools.co.uk/)
-- Navigate to Services > Management & Governance > Cloudformation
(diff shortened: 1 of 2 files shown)


lambda/README.md
```diff
@@ -10,20 +10,34 @@ The configuration for this application is stored in the [aws secret manager](htt
...
+  - Click 'Save'
+- Or use the AWS CLI
+  - Get the 'membership' aws credentials from [janus](https://janus.gutools.co.uk/) and add them to your local environment
+  - Get the existing configuration 
+  - get the current secrets 
+    ```bash
+    aws --region eu-west-1 --profile membership secretsmanager get-secret-value --version-stage AWSCURRENT --secret-id price-migration-engine-lambda-<STAGE>
+    ```
+  - The existing string is returned in the "SecretString" element of the json response, you will need to json string unescape
+    this value, and then make the changes/addtions to the result
```
kelvin-chappell, line 29:
```suggestion
    this value, and then make the changes/additions to the result
```

lambda/README.md
```diff
@@ -10,20 +10,34 @@ The configuration for this application is stored in the [aws secret manager](htt
...
+  - Create a new version of the secrets with the new secret string:
+    ```$bash
+    aws --region eu-west-1 --profile membership secretsmanager update-secret --secret-id price-migration-engine-lambda-<STAGE> --secret-string  '{"zuoraApiHost":"http://rest.apisandbox.zuora.com","zuoraClientId":"xxx","zuoraClientSecret":"xxx"}'
+    ```      
+- Update the secrect version in the cloudformation templates. The cloudformation templates contains mappings for the
```
kelvin-chappell, line 34:
```suggestion
- Update the secret version in the cloudformation templates. The cloudformation templates contains mappings for the
```

PR #25, opened 2020-05-18
Create skeleton salesforce price rise lambda 

frj:
This PR contains the skeleton code to create a new handler intended to create the Price_Rise__c objects in salesforce related to a particular price rise.

This includes:
- Cloudformation and riff-raff config for the lambda
- Stubbed out version of the Lambda handler class and new repository functions

### DIFF
--- a/lambda/cfn.yaml
+++ b/lambda/cfn.yaml
@@ -56,12 +56,48 @@ Resources:
                   - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/PriceMigrationEngine${Stage}"
                   - !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/PriceMigrationEngine${Stage}/*"
 
-  PriceMigrationEngineLambda:
+  PriceMigrationEngineEstimationLambda:
     Type: AWS::Lambda::Function
     Properties:
-      Description: Lambda used to implement the various task need in the process of increasing the price of subscriptions
+      Description: Lambda used to create estimated price, start date and other details of a price rise
       FunctionName:
-        !Sub price-migration-engine-lambda-${Stage}
+        !Sub price-migration-engine-estimation-lambda-${Stage}
+      Code:
+        S3Bucket: membership-dist
+        S3Key: !Sub membership/${Stage}/price-migration-engine-lambda/price-migration-engine-lambda.jar
+      Handler: pricemigrationengine.handlers.EstimationHandler::handleRequest
+      Environment:
+        Variables:
+          stage: !Ref Stage
+          zuoraApiHost:
+            !Sub
+              - '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:zuoraApiHost::${SecretsVersion}}}'
+              - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
+          zuoraClientId:
+            !Sub
+              - '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:zuoraClientId::${SecretsVersion}}}'
+              - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
+          zuoraClientSecret:
+            !Sub
+              - '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:zuoraClientSecret::${SecretsVersion}}}'
+              - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
+          earliestStartDate: 2020-05-15
+          batchSize: 100
+      Role:
+        Fn::GetAtt:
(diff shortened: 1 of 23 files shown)


lambda/cfn.yaml
```diff
@@ -92,3 +92,39 @@ Resources:
...
+              - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
+          zuoraClientId:
+            !Sub
+              - '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:zuoraClientId::${SecretsVersion}}}'
+              - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
+          zuoraClientSecret:
+            !Sub
+              - '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:zuoraClientSecret::${SecretsVersion}}}'
+              - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
+          earliestStartDate: 2020-05-15
```
kelvin-chappell, line 120: I don't think any of this will be needed

lambda/src/main/scala/pricemigrationengine/handlers/SalesforcePriceRiseCreationHandler.scala
```diff
@@ -0,0 +1,78 @@
...
+        .tapBoth(
+          e => Logging.error(s"Failed to write create Price_Rise in salesforce: $e"),
+          result => Logging.info(s"Estimated result: $result")
+        )
+      _ <- CohortTable
+        .update(result)
+        .tapBoth(
+          e => Logging.error(s"Failed to update Cohort table: $e"),
+          _ => Logging.info(s"Wrote $result to Cohort table")
+        )
```
kelvin-chappell, line 36: Perhaps update SF and then cohort table for each item

lambda/src/main/scala/pricemigrationengine/handlers/SalesforcePriceRiseCreationHandler.scala
```diff
@@ -0,0 +1,78 @@
...
+  ): ZLayer[Any, Any, Logging with Configuration with CohortTable with Zuora] = {
+    val configLayer = EnvConfiguration.impl
+    val cohortTableLayer =
+      loggingLayer ++ configLayer >>>
+        DynamoDBClient.dynamoDB ++ loggingLayer ++ configLayer >>>
+        DynamoDBZIOLive.impl ++ loggingLayer ++ configLayer >>>
+        CohortTableLive.impl
+    val zuoraLayer =
+      configLayer ++ loggingLayer >>>
+        ZuoraLive.impl
```
kelvin-chappell, line 56: No need for Zuora layer

lambda/src/main/scala/pricemigrationengine/services/CohortTableTest.scala
```diff
@@ -25,6 +26,8 @@ object CohortTableTest {
           for {
             _ <- console.putStrLn(s"Updating cohort table with result: $result")
           } yield ()
-    }
+
+        def update(result: SalesforcePriceRiseCreationResult): ZIO[Any, CohortUpdateFailure, Unit] = ???
+      }
   )
 }
```
kelvin-chappell, line 33: We can probably get rid of the `CohortTableTest` or move it to the test codebase

lambda/cfn.yaml
frj: I will have to create separate Config classes and config services for each of the lambdas if we dont want to use the same config for each app.

Which is probably a good thing. Ill do it

lambda/src/main/scala/pricemigrationengine/handlers/SalesforcePriceRiseCreationHandler.scala
frj: Actually, having given it a thumbs up, sorry I dont follow?

kelvin-chappell: Sorry, I misread this.  Ignore!

lambda/cfn.yaml
frj: It turns out to be a bit of a headache to have different config for each handler.

This is because of the shared services eg the DynamoDbLive class is dependent on Configuration, if I create a new config class eg SalesforcePriceRiseConfiguration then I cant wire the DynamoDbLive service into the handler that uses that configuration.

I think the best way would be to create completely separate config case classes/loader services for each of the service ie Zoura/Salesforce/DynamoDB and one each for the handlers with any of the remaining config items that are specific to that handler in it eg earliestStartDate

frj: Ok that is done, there is quite a bit of change in this PR now, but the config is much cleaner.

kelvin-chappell: Great work!

PR #29, opened 2020-05-20
Implement Zuora subscription update method

kelvin-chappell:


project/Dependencies.scala
```diff
@@ -1,9 +1,9 @@
 import sbt._
 
 object Dependencies {
-  private val zioVersion = "1.0.0-RC19"
+  private val zioVersion = "1.0.0-RC19-2"
 
-  lazy val awsDynamoDb = "com.amazonaws" % "aws-java-sdk-dynamodb" % "1.11.784"
+  lazy val awsDynamoDb = "com.amazonaws" % "aws-java-sdk-dynamodb" % "1.11.786"
```
frj, line 6: Very studious dependency management :)

kelvin-chappell: They seem to make a release roughly every half an hour but I'm keeping up

PR #32, opened 2020-05-21
Create price rise in salesforce

frj:
This PR contains the creation of the PriceRise__c object in salesforce during the execution of the SalesForcePriceRiseCreationHandler phase of the Price Migration Engine.

This includes:
- Building the price rise object out of the salesforce subscrption object and existing fields in the CohortTable.
- Implementation of the createPriceRise method in the Salesforce client
- Implementation of the CohortTable update method that writes the price rise id back to the table

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
+++ b/lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
@@ -1,7 +1,7 @@
 package pricemigrationengine.services
 
 import java.time.format.DateTimeFormatter
-import java.time.{Instant, LocalDate}
+import java.time.{Instant, LocalDate, ZoneOffset}
 import java.util
 
 import com.amazonaws.services.dynamodbv2.model.{AttributeAction, AttributeValue, AttributeValueUpdate, QueryRequest}
@@ -48,7 +48,8 @@ object CohortTableLive {
     estimationResult =>
       Map(
         stringFieldUpdate("processingStage", SalesforcePriceRiceCreationComplete.value),
-        stringFieldUpdate("salesforcePriceRiseId", estimationResult.id)
+        stringFieldUpdate("salesforcePriceRiseId", estimationResult.id),
+        instantFieldUpdate("whenSfShowEstimate", estimationResult.whenSfShowEstimate)
       ).asJava
 
   private implicit val cohortTableKeySerialiser: DynamoDBSerialiser[CohortTableKey] =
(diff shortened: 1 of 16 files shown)


lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
```diff
@@ -44,23 +64,62 @@ object CohortTableLive {
       AttributeAction.PUT
     )
 
+  private def instantFieldUpdate(fieldName: String, instant: Instant) =
+    fieldName -> new AttributeValueUpdate(
+      new AttributeValue().withS(DateTimeFormatter.ISO_DATE_TIME.format(instant.atZone(ZoneOffset.UTC))),
```
kelvin-chappell, line 69: [Instant.now](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#now--) gives the current UTC time, but could also use the ZIO `clock` as part of the environment as you're already using it.

lambda/src/main/scala/pricemigrationengine/services/CohortTableTest.scala
```diff
@@ -26,7 +25,7 @@ object CohortTableTest {
             _ <- console.putStrLn(s"Updating cohort table with result: $result")
           } yield ()
 
-        def update(result: SalesforcePriceRiseCreationResult): ZIO[Any, CohortUpdateFailure, Unit] = ???
+        def update(subscriptionName: String, result: SalesforcePriceRiseCreationDetails): ZIO[Any, CohortUpdateFailure, Unit] = ???
```
kelvin-chappell, line 28: `subscriptionNumber` seems to be the most common name, but it's not consistent in Zuora so probably interchangeable.

lambda/src/main/scala/pricemigrationengine/services/CohortTableTest.scala
```diff
@@ -1,7 +1,6 @@
 package pricemigrationengine.services
 
-import pricemigrationengine.handlers.SalesforcePriceRiseCreationResult
-import pricemigrationengine.model._
+import pricemigrationengine.model.{SalesforcePriceRiseCreationDetails, _}
```
kelvin-chappell, line 3:
```suggestion
import pricemigrationengine.model._
```

lambda/src/main/scala/pricemigrationengine/services/CohortTableTest.scala
frj: I was using subscriptionName based on that is what is used in the schema spreadsheet, we should probably make it consistent. Ill leave it as is we can come to a decision  and change it everywhere.

lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
frj: The instant does ultimately come from the ZIO.clock it gets passed in here via the cases class that is being serialised?

kelvin-chappell: Ignore!

PR #34, opened 2020-05-22
Add deployment config for amendment lambda

kelvin-chappell:


frj:
Good stuff, I was just being lazy just having one role :)

PR #46, opened 2020-06-01
Spread earliest start date over 3 months for monthly subs

kelvin-chappell:
There's a business requirement that the start date for monthlies is spread out over a quarter.
This is done by adding a random number of months to the earliest start date.


lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -38,9 +41,38 @@ object EstimationHandler extends App with RequestHandler[Unit, Unit] {
...
+   * Earliest start date spread out over 3 months.
+   */
+  private[handlers] def spreadEarliestStartDate(
+      subscription: ZuoraSubscription,
+      invoicePreview: ZuoraInvoiceList
+  ): ZIO[AmendmentConfiguration with Random, ConfigurationFailure, LocalDate] = {
+
+    def earliestStartDateForAMonthlySub(config: AmendmentConfig) =
+      for {
+        randomFactor <- random.nextIntBetween(0, 3)
```
frj, line 54: I think this will spread the subs over the next 4 months?

kelvin-chappell: The max value is exclusive: https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/random/package.scala#L111

frj: Ah sorry :)

PR #55, opened 2020-06-02
Tweaks for cases found in the live data so far

kelvin-chappell:


lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -158,11 +158,50 @@ object AmendmentData {
...
+        case Some("Quarter")     => Right(3)
+        case Some("Semi_Annual") => Right(6)
+        case Some("Annual")      => Right(12)
+        case other =>
+          Left(AmendmentDataFailure(s"Rate plan charge '${ratePlanCharge.number}' has unknown billing period: $other"))
+      }
+      ratePlanChargeAndPricing.pricing.price.filter(_ > 0) match {
+        case None => Right(None)
+        case Some(price) =>
+          multiplier(ratePlanChargeAndPricing.ratePlanCharge) map (multiplier => Some(price * multiplier))
```
frj, line 184: I am probably just missing something but isnt all of the pricing already in terms of the billing period of the subscription?

Ie is this multiplier needed?

PR #56, opened 2020-06-03
Read subscriptions to send emails from cohort table

frj:
This PR starts the implementation of the EmailNotificationHandler by reading CohortItems that are in the AmendmentComplete state and a startDate of 30 days in the future or less.

This includes:
- To support the "date range" element of the CohortTable query the ProcessingStageIndex has been changed to include startDate as a 'Range' Key. See:https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html
- Adding test for EmailNotificationHandler
- Improvements to the dynamoDB docs


lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
```diff
@@ -195,23 +195,34 @@ object CohortTableLive {
...
               cohortTableConfig <- CohortTableConfiguration.cohortTableConfig
                 .mapError(error => CohortFetchFailure(s"Failed to get configuration:${error.reason}"))
               stageConfig <- StageConfiguration.stageConfig
                 .mapError(error => CohortFetchFailure(s"Failed to get configuration:${error.reason}"))
+              queryRequest = new QueryRequest()
+                .withTableName(s"PriceMigrationEngine${stageConfig.stage}")
+                .withIndexName("ProcessingStageIndexV3")
+                .withKeyConditionExpression(
+                  "processingStage = :processingStage" + latestStartDateInclusive.fold("") { _ =>
+                    " AND startDate <= :latestStartDateInclusive"
```
kelvin-chappell, line 211: I thought we have to give *at least* 30 days' notice so shouldn't this be sending out `>=`?

frj: If I understand your point correctly, that wasn't my understanding of the requirements.

I had thought we would run the job every day, and send out notifications for items that had a start date in 30 days. The 'less than' is just there to make sure we hoover up any items that were missed for any reason.

If there are strict rules around the dm having to be send out no less than 30 days ahead of the start date, we need to take into account details Latcham lead time and batching behaviour. We can do that once we know what they are.

Ill double check with Sabrina what the rules are.

kelvin-chappell: 👍 
In that case this is good to go

frj: Yes so I checked with Sabrina and I think this is the correct behaviour, pending some tweaking around with the lead times.

PR #65, opened 2020-06-05
Get notification email data from salesforce

frj:
This PR adds the fetching fields from salesforce contact objects required to populate the price rise notification email.

These complete list of fields is:
contactID
IdentityID
First Name
Last Name
Address Line 1
Town
Postcode
County
Country
New price
Next payment date
Frequency of bill 

This includes:
- Implementing getContact in the salesforce client
- Mapping of CohortItem and Contact fields to the EmailMessage case class, this is the format of the sqs message that will be sent to membership workflow.
- Creating a no-op logging implementation of the EmailSenderService


lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
```diff
@@ -11,27 +12,71 @@ import zio.{Runtime, ZEnv, ZIO, ZLayer, clock, console}
 object NotificationEmailHandler {
   private val NotificationEmailLeadTimeDays = 30
```
kelvin-chappell, line 13: Perhaps should go in config

lambda/src/main/scala/pricemigrationengine/model/Failure.scala
```diff
@@ -23,3 +23,4 @@ case class S3Failure(reason: String) extends Failure
 case class SubscriptionIdUploadFailure(reason: String) extends Failure
 
 case class NotificationEmailFailure(reason: String) extends Failure
+case class EmailSenderFailure(reason: String) extends Failure
```
kelvin-chappell, line 26: What's the difference between these?

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
frj: I thought about that but you need to make a code change to change the config version in the cloudformation template to get any change to the config to take effect. 

So it seems easier to just have the config in code. Its a drawbacks of using cloudformation dynamic references for config.

lambda/src/main/scala/pricemigrationengine/model/Failure.scala
frj: Oops

frj: Ah, actually they are different, one is raised by the EmailSender service one by the NotificationEmailHandler.

Ill leave it as is.

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
kelvin-chappell: 👍

PR #71, opened 2020-06-11
Create subscription update with charge overrides

kelvin-chappell:
When the billing period of a rate plan charge is different to the billing period of its corresponding product rate plan charge, we have add a charge override to the subscription update to reflect the different price over a different period.
In most cases this won't be relevant, but with some products there's a base monthly price and for longer periods the price is just a multiple of the base price.

So, for matching billing periods, update looks like:
```
{
  "add": [
    {
      "productRatePlanId": "prp1",
      "contractEffectiveDate": "2020-06-01"
    }
  ],
  "remove": [
    {
      "ratePlanId": "rp1",
      "contractEffectiveDate": "2020-06-01"
    }
  ]
}
```

And for different billing periods:
```
{
  "add": [
    {
      "productRatePlanId": "prp1",
      "contractEffectiveDate": "2020-06-01",
      "chargeOverrides": [
        {
          "productRatePlanChargeId": "prpc1",
          "billingPeriod": "Quarter",
          "price": 1.23
        },
        {
          "productRatePlanChargeId": "prpc2",
          "billingPeriod": "Quarter",
          "price": 2.34
        }
      ]
    }
  ],
  "remove": [
    {
      "ratePlanId": "rp1",
      "contractEffectiveDate": "2020-06-01"
    }
  ]
}
```

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/model/Either.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/Either.scala
@@ -1,0 +1,15 @@
+package pricemigrationengine.model
+
+object Either {
+
+  /**
+    * Converts a Seq of Eithers to either a Left
+    * containing the first Left value of the Seq if any of the Seq has a Left value,
+    * or else to a Right that contains the Seq of all the Right values in the original Seq.
+    */
+  def fromSeq[E, A](eithers: Seq[Either[E, A]]): Either[E, Seq[A]] =
+    eithers
+      .collectFirst { case Left(e) => e }
+      .map(Left(_))
+      .getOrElse(Right(eithers.collect { case Right(a) => a }))
+}
--- a/lambda/src/main/scala/pricemigrationengine/model/Either.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/Either.scala
@@ -7,9 +7,12 @@ object Either {
     * containing the first Left value of the Seq if any of the Seq has a Left value,
     * or else to a Right that contains the Seq of all the Right values in the original Seq.
     */
-  def fromSeq[E, A](eithers: Seq[Either[E, A]]): Either[E, Seq[A]] =
-    eithers
-      .collectFirst { case Left(e) => e }
-      .map(Left(_))
-      .getOrElse(Right(eithers.collect { case Right(a) => a }))
+  // to keep the lambda package small we are not importing cats
+  implicit final class TraverseOps[E, A](eithers: Seq[Either[E, A]]) {
+    def sequence: Either[E, Seq[A]] =
+      eithers
+        .collectFirst { case Left(e) => e }
+        .map(Left(_))
+        .getOrElse(Right(eithers.collect { case Right(a) => a }))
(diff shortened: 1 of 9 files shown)


lambda/src/main/scala/pricemigrationengine/model/Either.scala
```diff
@@ -0,0 +1,15 @@
+package pricemigrationengine.model
+
+object Either {
+
+  /**
+    * Converts a Seq of Eithers to either a Left
+    * containing the first Left value of the Seq if any of the Seq has a Left value,
+    * or else to a Right that contains the Seq of all the Right values in the original Seq.
+    */
+  def fromSeq[E, A](eithers: Seq[Either[E, A]]): Either[E, Seq[A]] =
```
mario-galic, line 10: Are you trying to keep the lambda size small? For familiarity you could call it `sequence` and provide it as extension method

```
// to keep the lambda package small we are not importing cats
implicit final class TraverseOps[E, A](eithers: Seq[Either[E, A]]) {
  def sequence: Either[E, Seq[A]] = {
    eithers
      .collectFirst { case Left(e) => e }
      .map(Left(_))
      .getOrElse(Right(eithers.collect { case Right(a) => a }))
  }
}
```

so now you can use the familiar `l.sequence` syntax.

lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscriptionUpdate.scala
```diff
@@ -34,22 +39,30 @@ object ZuoraSubscriptionUpdate {
     if (ratePlans.isEmpty)
       Left(AmendmentDataFailure("No rate plans to update"))
     else
-      Right(
-        ZuoraSubscriptionUpdate(
-          add = ratePlans.map(ratePlan => AddZuoraRatePlan(ratePlan.productRatePlanId, date)),
-          remove = ratePlans.map(ratePlan => RemoveZuoraRatePlan(ratePlan.id, date))
-        )
+      for {
```
mario-galic, line 42: No need for for-comprehension

```
Either
  .fromSeq(ratePlans.map(AddZuoraRatePlan.fromRatePlan(pricingData, date)))
  .map(ZuoraSubscriptionUpdate(_, ratePlans.map(ratePlan => RemoveZuoraRatePlan(ratePlan.id, date))))
```

lambda/src/main/scala/pricemigrationengine/model/Either.scala
kelvin-chappell: That's a good idea.  I'll do it.  I didn't want to bring in a whole library for this although I had a feeling it was done elsewhere.

lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscriptionUpdate.scala
kelvin-chappell: Yes that's better.

PR #72, opened 2020-06-15
Cloudform initial state machine

kelvin-chappell:
This change includes the [cloudformation template](https://github.com/guardian/price-migration-engine/blob/51d5feefc2780520450ad16401de9510aaa2708d/stateMachine/cfn.yaml) for a state machine that's already been used in production:

stepfunctions_graph

Eventually this will hold all the steps involved in a price migration.

When a lambda times out, it returns a [Lambda.Unknown](https://github.com/guardian/price-migration-engine/compare/kc-state-machine?expand=1#diff-bd341ba6709387aedaa0b70e5607e942R94) error.  So we take this error as a signal that there is more work to do and re-run the lambda.  Once it completes within the time available to the lambda, the step is considered to be successful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,4 +3,7 @@
 This repo consists of a set of lambdas and DynamoDB table templates that are designed to interact together in a state 
 machine.  
 
-Documentation for the lambda code is in the [lambda readme doc](lambda/README.md).
+See readme for:  
+* [State machine](stateMachine/README.md)
+* [Lambdas](lambda/README.md)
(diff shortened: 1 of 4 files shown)


kelvin-chappell:
@frj Yes it would need to run multiple times.  Haven't quite worked out best way to do it yet.  To be honest I was really just seeing what state machines can do.  In practice, it might be better just to do all the estimations and then write them to SF.

PR #78, opened 2020-06-17
Update cohort item status for email send

frj:
This PR add the updating of the CohortTable processingStage for the completion of the NotificationEmailHandler.

The processing of sending emails is not idempotent, ie its possible the sending of the notification via SQS might return failure eg due to network failure, but the message may have been sent successfully. 

To avoid resending direct mails to the customer, the CohortItem will be put into a NotificationEmailProcessing before sending the notification. 

In the failure case the item will stay in that state, requiring manual intervention to put it into a state where it will continue to be processed.


lambda/README.md
```diff
@@ -34,6 +34,41 @@ so that the artefact generated is of minimal size and lambdas can warm up quickl
...
+
+| Processing stage at start | Lambda | Description | Processing stage on completion |
+|---------|---|---|---|
+| N/A | SubscriptionIdUploadHandler | Initialises the items in the CohortTable for more details see:See [ImportSubscriptionId.MD](ImportSubscriptionId.MD) | ReadyForEstimation |
+| ReadyForEstimation | EstimationHandler | Uses Zuora to 'estimate' new price and start date of price rise | EstimationComplete |
+| EstimationComplete | SalesforcePriceRiseCreationHandler | Creates the prices rise object in SF so the estimated information is available to CSRs | SalesforcePriceRiceCreationComplete/Cancelled (if cancellation detected) |
+| SalesforcePriceRiceCreationComplete | NotificationEmailHandler | Sends prices rise notification direct mail to customer via braze | EmailSendProcessing (on failure)/EmailSendComplete (on success) |
+| EmailSendComplete | AmendmentHandler | Applies the prices rise amendment to Zuora | AmendmentComplete/Cancelled (if cancellation detected) |
+ 
+
```
kelvin-chappell, line 71: 💯

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
```diff
@@ -58,9 +60,9 @@ object NotificationEmailHandler {
                 billing_postal_code = postalCode,
                 billing_state = contact.MailingAddress.state,
                 billing_country = country,
-                payment_amount = newPrice,
+                payment_amount = estimatedNewPrice,
                 next_payment_date = startDate,
-                payment_frequency = billingPeriod,
+                payment_frequency = s"${billingPeriod}ly",
```
kelvin-chappell, line 65: I think there's a `Semi_Annual` billing period, but will need to check that.  If so, might be better formatted as `semiannually`.

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
```diff
@@ -70,13 +72,30 @@ object NotificationEmailHandler {
...
     } yield ()
   }
 
   def requiredField(field: Option[String], fieldName: String): ZIO[Any, NotificationEmailHandlerFailure, String] = {
     ZIO.fromOption(field).orElseFail(NotificationEmailHandlerFailure(s"$fieldName is a required field"))
   }
 
+  def updateCohortItemStatus(subscriptionNumber: String, processingStage: CohortTableFilter) = {
+    for {
+      now <- clock.currentDateTime.mapError(ex => NotificationEmailHandlerFailure(s"Failed to get time: $ex"))
```
kelvin-chappell, line 86: Is it possible not to get current time?

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
```diff
@@ -39,10 +39,12 @@ object NotificationEmailHandler {
       street <- requiredField(contact.MailingAddress.street, "Contact.MailingAddress.street")
       postalCode <- requiredField(contact.MailingAddress.postalCode, "Contact.MailingAddress.postalCode")
       country <- requiredField(contact.MailingAddress.country, "Contact.MailingAddress.country")
-      newPrice <- requiredField(cohortItem.newPrice.map(_.toString()), "CohortItem.newPrice")
+      estimatedNewPrice <- requiredField(cohortItem.estimatedNewPrice.map(_.toString()), "CohortItem.estimatedNewPrice")
       startDate <- requiredField(cohortItem.startDate.map(_.toString()), "CohortItem.startDate")
       billingPeriod <- requiredField(cohortItem.billingPeriod, "CohortItem.billingPeriod")
 
+      _ <- updateCohortItemStatus(cohortItem.subscriptionName, EmailSendProcessing)
```
kelvin-chappell, line 46: Instead of a `processing` state, would it be better to have a `complete` and a `failed` state?  If the message fails to go on the queue, could update state to `failed` which indicates it was tried.  

I'm just thinking that if you query the table at any moment to find all the messages that appear to have failed so that you could cross-refer them with what Braze says it sent, the `processing` state might be misleading as they haven't definitely failed at that moment.  I know that's quite an unlikely scenario.

kelvin-chappell: There is transaction support in Dynamo although it adds some complexity:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-example.html so could read and write back state in a single transaction.

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
frj: Are you thinking of passing it through from the top level?

I guess it wouldnt be quite as accurate, the email job may take some time to run, so the time stamp could be minute out of sync with when the email is actually sent.

I guess its not a big deal but whats the object to getting the time here?

kelvin-chappell: I was just commenting on the exception mapping.  I wouldn't think `clock.currentDateTime` could fail.

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
kelvin-chappell: Ignore this comment if too complicated.  Solution is fine.

frj: I think there are lots of obscure failure cases eg if dynamoDB goes down or is not contactable during email run. 

In which case it wouldnt be possible to set dynamoDB to a 'Failed' state.

Conversely a network problem could cause the SQS send to look like a failure when the email was actually sent, so in that case setting the status to a Failed state would be miss-leading.

In any failure case the action is the same ie to go and reconcile the items still in the 'processing' state with braze/membership-workflow and explisitly set

kelvin-chappell: 👍

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
frj: Ill put in a mapping and fail if we see any values not in the mapping table.

kelvin-chappell: 👍

lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
frj: Ah, well it is a side-effect, you never know :)

The clock is the Zio clock as such you have to handle the failure case. 

You are right it might be a bit unnecessary but Im using it to stub out the clocks in the test and Ive used it fairly consistently in the code.

Ill leave it as is, maybe we can go through and change everything over to use java.time.Clock instead.

PR #84, opened 2020-06-19
logging count of price rise notifications sent by NotificationEmailHandler

frj:
Logging count of price rise notifications sent by NotificationEmailHandler

Also correcting example dynamodb queries in readme


lambda/src/main/scala/pricemigrationengine/handlers/NotificationEmailHandler.scala
```diff
@@ -23,7 +23,8 @@ object NotificationEmailHandler {
         SalesforcePriceRiceCreationComplete,
         Some(now.toLocalDate.plusDays(NotificationEmailLeadTimeDays))
       )
-      _ <- subscriptions.foreach(sendEmail)
+      count <- subscriptions.mapM(sendEmail).runCount
```
kelvin-chappell, line 26: Very useful! 💯

PR #85, opened 2020-06-19
DynamoDb implementation of CohortSpecTable

kelvin-chappell:
This implements the `fetchAll` method.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/services/CohortSpecTableLive.scala
+++ b/lambda/src/main/scala/pricemigrationengine/services/CohortSpecTableLive.scala
@@ -1,0 +1,63 @@
+package pricemigrationengine.services
+
+import java.time.LocalDate
+
+import com.amazonaws.services.dynamodbv2.model.{AttributeValue, ScanRequest}
+import pricemigrationengine.model._
+import zio.{IO, ZIO, ZLayer}
+
+import scala.collection.mutable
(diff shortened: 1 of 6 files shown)


lambda/src/main/scala/pricemigrationengine/services/CohortSpecTableLive.scala
```diff
@@ -0,0 +1,63 @@
...
+      fields.get(fieldName).map(_.getS).toRight(CohortSpecFetchFailure(s"Can't read field '$fieldName'"))
+
+    def optStringFieldValue(fields: mutable.Map[String, AttributeValue], fieldName: String): Option[String] =
+      fields.get(fieldName).map(_.getS)
+
+    for {
+      cohortName <- stringFieldValue(values, "cohortName")
+      earliestPriceMigrationStartDate <- stringFieldValue(values, "earliestPriceMigrationStartDate").map(
+        LocalDate.parse)
+      importStartDate <- stringFieldValue(values, "importStartDate").map(LocalDate.parse)
```
frj, line 29: Should we explicitly handle the potential parsing failure here.

Also there are a bunch of these conversion functions in the CohortTableLive, maybe it would be worth moving them to shared code?

kelvin-chappell: Yes for both.  I'll update the PR

PR #86, opened 2020-06-22
Add ad-hoc status query

kelvin-chappell:
Query is designed so that its output can be fed into one of the ad-hoc update or delete scripts already written.

### DIFF
--- a/dynamoDb/scripts/cohortStateQuery.bash
+++ b/dynamoDb/scripts/cohortStateQuery.bash
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# =====================================================================
+# Gives a line-separated list of subscription numbers
+# that are in a given processing stage
+# and have a start date up to and including a given threshold.
+#
+# Arg 1: Deployment stage: DEV, CODE or PROD
+# Arg 2: processingStage, eg. EstimationComplete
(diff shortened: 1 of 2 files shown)


dynamoDb/scripts/cohortStateQuery.bash
```diff
@@ -0,0 +1,26 @@
...
+# Arg 2: processingStage
+# Arg 3: Inclusive maximum date
+function query() {
+  aws dynamodb query --region eu-west-1 --profile membership \
+    --table-name "PriceMigrationEngine$1" \
+    --index-name "ProcessingStageIndexV3" \
+    --projection-expression "subscriptionNumber" \
+    --key-condition-expression "processingStage = :s AND startDate <= :d " \
+    --expression-attribute-values "{\":s\": {\"S\": \"$2\"}, \":d\": {\"S\": \"$3\"}}" \
+    --page-size "100000"
```
frj, line 23: I think this should be using the --max-items, I think page-size is the internal internal thing where as --max-items set the number of values returned by the query ie the actual page size :)

kelvin-chappell: I see. I read `max-items` as limit.  I'll try that instead.

PR #88, opened 2020-06-22
Write mail sent timestamp to sf

frj:
This PR creates a handler that writes the date on which the prices rise notification mail was sent for a particular subscription into salesforce, allowing the CSRs to inform customers when they were notified about a price rise.

The handler follows the process:
- Read CohortItems with a status of 'EmailSendComplete'
- Convert the CohortItem.whenEmailSend into a Date
- Write the date to the salesforce Price_Rise__c object using a PATCH/update request
- Update the CohortItem setting the processingStage to 'EmailSendDateWrittenToSalesforce' and the 'whenEmailSentWrittenToSalesforce' field to the current time.

Other changes:
- Extracted test logging duplicated code into shared class
- Added test for salesforce request body serialisation to ensure correct behaviour when updating SF

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceNotificationDateUpdateHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceNotificationDateUpdateHandler.scala
@@ -29,18 +29,18 @@ object SalesforceNotificationDateUpdateHandler extends App with RequestHandler[U
         )
       time <- clock.currentDateTime
         .mapError { error =>
-          SalesforcePriceRiseCreationFailure(s"Failed to get currentTime: $error")
+          SalesforcePriceRiseWriteFailure(s"Failed to get currentTime: $error")
         }
-      salesforcePriceRiseCreationDetails = CohortItem(
+      salesforcePriceRiseDetails = CohortItem(
         subscriptionName = item.subscriptionName,
         processingStage = EmailSendDateWrittenToSalesforce,
         whenEmailSentWrittenToSalesforce = Some(time.toInstant)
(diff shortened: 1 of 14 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/SalesforceNotificationDateUpdateHandler.scala
```diff
@@ -0,0 +1,104 @@
...
+  ): ZIO[Logging with CohortTable with SalesforceClient with Clock, Failure, Unit] =
+    for {
+      _ <- updateSalesforce(item)
+        .tapBoth(
+          e => Logging.error(s"Failed to write create Price_Rise in salesforce: $e"),
+          result => Logging.info(s"SalesforcePriceRise result: $result")
+        )
+      time <- clock.currentDateTime
+        .mapError { error =>
+          SalesforcePriceRiseCreationFailure(s"Failed to get currentTime: $error")
```
kelvin-chappell, line 32: Perhaps should be renamed as `SalesforcePriceRiseWriteFailure` now that its use is being extended

lambda/src/main/scala/pricemigrationengine/handlers/SalesforceNotificationDateUpdateHandler.scala
```diff
@@ -0,0 +1,104 @@
...
+        }
+      salesforcePriceRiseCreationDetails = CohortItem(
```
kelvin-chappell, line 34:
```suggestion
      salesforcePriceRiseDetails = CohortItem(
```

PR #109, opened 2020-07-01
Extend subscription term when necessary

kelvin-chappell:
We can only apply an amendment to a subscription if the effective date of the amendment isn't later than the current term-end date of the subscription.
So this change extends the current term of the subscription to include the effective date of the amendment.


frj:
That old chessnut

PR #116, opened 2020-07-06
Implement CohortStateMachine service

kelvin-chappell:
This wires together the [MigrationHandler](https://github.com/guardian/price-migration-engine/blob/kc-machine-kick/lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala) with the [CohortStateMachineLive](https://github.com/guardian/price-migration-engine/blob/kc-machine-kick/lambda/src/main/scala/pricemigrationengine/services/CohortStateMachineLive.scala) implementation.

The [CohortStateMachineConfig](https://github.com/guardian/price-migration-engine/compare/kc-machine-kick?expand=1#diff-48fde3af0d999d859f698d02ce587fb0R35) gives the ARN of the state machine.  Which cohort the state machine actually operates on depends on the content of the [CohortSpec](https://github.com/guardian/price-migration-engine/blob/kc-machine-kick/lambda/src/main/scala/pricemigrationengine/model/CohortSpec.scala).

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/services/CohortStateMachineLive.scala
+++ b/lambda/src/main/scala/pricemigrationengine/services/CohortStateMachineLive.scala
@@ -1,0 +1,39 @@
+package pricemigrationengine.services
+
+import java.time.LocalDate
+
+import com.amazonaws.regions.Regions.EU_WEST_1
+import com.amazonaws.services.stepfunctions.AWSStepFunctionsClientBuilder
+import com.amazonaws.services.stepfunctions.model.{StartExecutionRequest, StartExecutionResult}
+import pricemigrationengine.model._
+import upickle.default.write
+import zio.{IO, ZIO, ZLayer}
+
+object CohortStateMachineLive {
+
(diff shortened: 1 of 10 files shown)


lambda/src/main/scala/pricemigrationengine/services/CohortStateMachineLive.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    ZLayer.fromServicesM[CohortStateMachineConfiguration.Service,
+                         Logging.Service,
+                         Any,
+                         ConfigurationFailure,
+                         CohortStateMachine.Service] { (configuration, logging) =>
+      configuration.config map { config =>
+        new CohortStateMachine.Service {
+          def startExecution(date: LocalDate)(spec: CohortSpec): IO[CohortStateMachineFailure, StartExecutionResult] =
+            ZIO
```
frj, line 26: I keep meaning to start using effectBlocking for all of these calls to the aws client codes. Its not a problem as there is not really any parallelism, I might go through and fix it at some point.

kelvin-chappell: That would be an improvement.  I'll add it in this case.

PR #122, opened 2020-07-13
removing limit of number of notifications sent byt NotificationHandler

frj:
Processing all price rise notification letters due to be processed.


frj:
Yes so far the 15min limit hasn't been a problem

PR #123, opened 2020-07-13
Repeat each step in state machine until complete

kelvin-chappell:
This gives each step a value to specify if it's complete, which the state machine then uses to decide if to repeat the step.
This is the same as has been done for the `EstimatingLambda` in #120 but for all the other steps.

Also the input is no longer returned as part of the output of the lambdas as that's redundant.  It's passed through by default by the state machine.
See https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html#input-output-resultpath-append.

The state machine now looks like this:

stepfunctions_graph

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceAmendmentUpdateHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceAmendmentUpdateHandler.scala
@@ -30,7 +30,7 @@ object SalesforceAmendmentUpdateHandler extends CohortHandler {
           )
       itemsToGo <- fetchFromCohortTable
       numItemsToGo <- itemsToGo.take(1).runCount
-    } yield HandlerOutput(cohortSpec, isComplete = numItemsToGo == 0)
+    } yield HandlerOutput(isComplete = numItemsToGo == 0)
 
   private def fetchFromCohortTable = CohortTable.fetch(AmendmentComplete, None)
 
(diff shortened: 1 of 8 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/SalesforceAmendmentUpdateHandler.scala
```diff
@@ -1,46 +1,59 @@
...
+          .take(batchSize)
+          .foreach(item =>
+            updateSfWithNewSubscriptionId(item).tapBoth(
+              e => Logging.error(s"Failed to update price rise record for ${item.subscriptionName}: $e"),
+              _ => Logging.info(s"Amendment of ${item.subscriptionName} recorded in Salesforce")
+            )
+          )
+      itemsToGo <- fetchFromCohortTable
+      numItemsToGo <- itemsToGo.take(1).runCount
+    } yield HandlerOutput(isComplete = numItemsToGo == 0)
```
frj, line 33: Not sure if this is better, in some rare cases the lambda might get re-executed unnecessarily, but I will save re-querying the table to check for more items.  

```
      count <-
        cohortItems
          .take(batchSize)
          .map(item =>
            updateSfWithNewSubscriptionId(item).tapBoth(
              e => Logging.error(s"Failed to update price rise record for ${item.subscriptionName}: $e"),
              _ => Logging.info(s"Amendment of ${item.subscriptionName} recorded in Salesforce")
            )
          )
          .runCount
    } yield HandlerOutput(isComplete = count < batchSize)
```

kelvin-chappell: That is better!  Thanks @frj

PR #127, opened 2020-07-15
Add step to state machine to update SF after notifications

kelvin-chappell:
Machine now looks like this:
stepfunctions_graph

### DIFF
--- a/lambda/cfn.yaml
+++ b/lambda/cfn.yaml
@@ -292,7 +292,7 @@ Resources:
   PriceMigrationEngineSalesforcePriceCreationLambda:
     Type: AWS::Lambda::Function
     Properties:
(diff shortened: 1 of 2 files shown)


lambda/cfn.yaml
```diff
@@ -292,7 +292,7 @@ Resources:
   PriceMigrationEngineSalesforcePriceCreationLambda:
     Type: AWS::Lambda::Function
     Properties:
-      Description: Lambda used to create Price_Rise__c objects in salesforce realting to the price rise.
+      Description: Lambda used to create Price_Rise__c objects in salesforce relating to the price rise.
```
kelvin-chappell, line 295: Have to make a change to a resource to get cloudformation to pick up any new exports.
ie. only changes in resources act as a trigger but then exports get detected as well.
Otherwise exports just get ignored.

PR #132, opened 2020-07-21
Fallback to mailing address if no billing address

kelvin-chappell:


lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -124,14 +126,12 @@ object NotificationHandler extends CohortHandler {
...
-      Logging.info(
+  def logMissingEmailAddress(cohortItem: CohortItem, sfContact: SalesforceContact): ZIO[Logging, Nothing, Unit] = {
+    Logging
+      .info(
         s"Subscription ${cohortItem.subscriptionName} is for contact ${sfContact.Id} that has not email address"
       )
-    } else {
-      ZIO.unit
-    }
+      .when(sfContact.Email.isEmpty)
```
frj, line 134: Nice :)

PR #136, opened 2020-07-23
Modify logic of recipient's title and first name

kelvin-chappell:
The letter template expects a first name.  So when there's no first name, we use the title as the first name and leave the title field empty.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,3 +1,3 @@
-version = 2.6.2
+version = 2.6.4
 
 maxColumn = 120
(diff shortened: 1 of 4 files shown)


frj:
I don't think salutation is a mandatory field so Im dont think this is going to fully get around the need to relax latchams validation for first name.

That being the case should we just get latchams to change the validation.

Or we could go with this until that problem comes up?

kelvin-chappell:
@frj In cases where SF has no first name and no title we're still stuck, which is why a neutral title would be good to use as the ultimate fallback.  But happy to wait until that happens.  Both of the cases that have come up so far have titles but there are a lot of contacts that don't have either first name or title.

PR #139, opened 2020-07-24
Export cohort table to s3

frj:
This PR contains the work in progress for the export of Cohort Items to s3 as a CSV file for consumption by the data lake etl pipeline.

The Handler is triggered in the normal way supplying the details of the Cohort to be exported as input json.

S3 requires a content length header at the beginning of the PUT request, this means the size of the file to be uploaded has to be determined before starting to write the file. This prevents streaming of CohortItems directly from dynamoDb to S3, requiring the file be buffered completely to determine its length.

The file can be buffered in two ways:
  - In Memory - This puts a hard limit on the size of file that can be written to s3
  - On disk - Lambdas have 500mb of /tmp file space however it is possible to mount an efs volume on the lambda which provides 'unlimited' storage

This implementation uses the on disk in the /tmp storage, the current cohort produces a CSV of size 13mb so we dont expect to exceed the 500mb limit, but the efs solution provides a way to support larger files.

Details:
  - Implemented CohortTable.fetchAll using dynamoDb scan operation this supports fetching the whole unfiltered dynamodb table with no indexes required
 
This change is WIP the following work is still required:
- Possibly export data from all 'Active' Cohort tables, this depends on the outcome of discussions with the data tech team.
- Move export file to 'Raw' s3 bucket, pending creation of the bucket.
- Tests
- Documentation
- Scheduling


lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
```diff
@@ -175,9 +175,27 @@ object CohortTableLive {
                     _ => Logging.info(s"Wrote $result to Cohort table")
                   )
               }.provide(dependencies)
+
+              override def fetchAll(): IO[CohortFetchFailure, ZStream[Any, CohortFetchFailure, CohortItem]] = {
+                for {
+                  cohortTableConfig <- CohortTableConfiguration.cohortTableConfig
+                    .mapError(error => CohortFetchFailure(s"Failed to get configuration:${error.reason}"))
+                  stageConfig <- StageConfiguration.stageConfig
```
kelvin-chappell, line 183: `StageConfig` is already available from the impl construction at [line 117](https://github.com/guardian/price-migration-engine/blob/98cac5036bbf612c574b6bdae00a85799b99880f/lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala#L117)

lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala
```diff
@@ -175,9 +175,27 @@ object CohortTableLive {
                     _ => Logging.info(s"Wrote $result to Cohort table")
                   )
               }.provide(dependencies)
+
+              override def fetchAll(): IO[CohortFetchFailure, ZStream[Any, CohortFetchFailure, CohortItem]] = {
+                for {
+                  cohortTableConfig <- CohortTableConfiguration.cohortTableConfig
```
kelvin-chappell, line 181: Could `cohortTableConfig` be brought in during the impl construction like `stageConfig` at [line 117](https://github.com/guardian/price-migration-engine/blob/98cac5036bbf612c574b6bdae00a85799b99880f/lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala#L117)?
Then it could give a `ConfigurationFailure` before trying to fetch.

lambda/src/main/scala/pricemigrationengine/handlers/CohortTableDatalakeExportHandler.scala
```diff
@@ -0,0 +1,158 @@
...
+        "whenNotificationSentWrittenToSalesforce",
+        "whenAmendmentWrittenToSalesforce"
+      )
+    ).use { printer =>
+      cohortItems.mapM { cohortItem =>
+        ZIO
+          .effect(
+            printer.printRecord(
+              cohortSpec.cohortName,
+              cohortSpec.tmpTableName.getOrElse(""),
```
kelvin-chappell, line 108: Perhaps
```suggestion
              cohortSpec.tableName.getOrElse(""),
```
would be better to get table name in every case

PR #142, opened 2020-07-27
Import subs safely

kelvin-chappell:
Previously we were 'putting' new subs into the table, which meant items would be overwritten if they were already present.

This PR adds a condition to the put method to make [a new safe 'create' method](https://github.com/guardian/price-migration-engine/compare/kc-import?expand=1#diff-052261b41151e8ed656f54b901dbb3c0R51-R54), which won't overwrite an existing item.
Then in order to handle failures partway through an import, we ignore [any cases where an item already exists in the table](https://github.com/guardian/price-migration-engine/compare/kc-import?expand=1#diff-e18cb17f719bc556c7a9480d39b35fe1R87-R90) and just log them instead.  This should make the import lambda safely re-runnable.

This assumes import can be done in 15 mins.  Otherwise will have to do some kind of comparison of what's already in table with what's to be imported.


kelvin-chappell:
I see what you mean - an update would either create or ignore.  Oh well, we have a special create method now!

PR #145, opened 2020-07-28
adding missing CohortTableFilter

frj:
Adding a 'missing' cohort table filter 'AmendmentFailed' and ensured the CohortFilter.all field is up to date.


lambda/src/main/scala/pricemigrationengine/model/CohortTableFilter.scala
```diff
@@ -24,6 +24,8 @@ object CohortTableFilter {
 
   case object AmendmentComplete extends CohortTableFilter { override val value: String = "AmendmentComplete" }
 
+  case object AmendmentFailed extends CohortTableFilter { override val value: String = "AmendmentFailed" }
```
kelvin-chappell, line 27: Should probably go below L42 with the exceptions

PR #160, opened 2020-08-28
Throttle send email

mario-galic:
Try to stay under 1000 Latcham webhook triggers per minute. Paul's performance testing demonstrated 500 response on above 1000 per minute, which means Braze would retry the webhook, which might be the cause of duplicate records in Latcham (TBC).

- [Braze retry logic](https://www.braze.com/docs/user_guide/message_building_by_channel/webhooks/creating_a_webhook/#errors-retry-logic-and-timeouts)
- [Latcham Direct - API test cases](https://docs.google.com/spreadsheets/d/1d_UpFB9JzkfRXfG_Rel2jUq3ORvYO3us6ru9i59vn0w/edit#gid=5475790)
- [Rates of webhook trigger on 2020-08-22](https://docs.google.com/spreadsheets/d/18WWCApCKn-MsUDOcYmtXx4wJyUcgHyaOSbmVnGOqU2c/edit?ts=5f47edbc#gid=989130158)

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -16,6 +16,7 @@ import pricemigrationengine.model.membershipworkflow.{
 import pricemigrationengine.services._
 import zio.clock.Clock
 import zio.{ZEnv, ZIO, ZLayer}
+import zio.duration._
 
 object NotificationHandler extends CohortHandler {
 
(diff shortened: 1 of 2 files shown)


paulbrown1982 (approved):
👍LGTM

kelvin-chappell:
Thanks for this @mario-galic.  I like the way the throttle works but it looks like the rate limit you set in Braze coupled with the Latcham dedup should do the trick.

Also, if this is likely to be a problem with any Braze webhook perhaps we should do the throttling logic in membership-workflow where the direct-mail queue ends up?  Or alternatively configure the queue itself to have a rate limit.

PR #164, opened 2020-09-04
Bump dependencies

kelvin-chappell:
This puts all dependencies on latest versions.

Bumping ZIO to v1.0.1 entailed changing argument types of `ZIO.foreach`.

Also removed all unused imports to please the compiler.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
@@ -16,7 +16,7 @@ object MigrationHandler extends zio.App with RequestHandler[Unit, Unit] {
       cohortSpecs <- CohortSpecTable.fetchAll
       activeSpecs <-
         ZIO
-          .filter(cohortSpecs)(cohort => ZIO.succeed(CohortSpec.isActive(cohort)(today)))
+          .filter(cohortSpecs.toList)(cohort => ZIO.succeed(CohortSpec.isActive(cohort)(today)))
           .tap(specs => Logging.info(s"Currently ${specs.size} active cohorts"))
       _ <- ZIO.foreach(activeSpecs)(CohortStateMachine.startExecution)
     } yield ()).tapError(e => Logging.error(s"Migration run failed: $e"))
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,8 @@
 import Dependencies._
 import sbt.Keys.{description, name}
 
+Global / onChangedBuildSource := ReloadOnSourceChanges
+
 ThisBuild / scalaVersion := "2.13.3"
 
 lazy val root = (project in file("."))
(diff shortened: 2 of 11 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
```diff
@@ -16,7 +16,7 @@ object MigrationHandler extends zio.App with RequestHandler[Unit, Unit] {
       cohortSpecs <- CohortSpecTable.fetchAll
       activeSpecs <-
         ZIO
-          .filter(cohortSpecs)(cohort => ZIO.succeed(CohortSpec.isActive(cohort)(today)))
+          .filter(cohortSpecs.toList)(cohort => ZIO.succeed(CohortSpec.isActive(cohort)(today)))
```
arausuy, line 19: No worries about having to do this extra conversion or is it forced due to an api change?

arausuy (approved):

lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
kelvin-chappell: Unfortunately it's forced by an API change between `1.0.0-RC21-2` and `1.0`.
The argument now has to be an `Iterable`, which seems a bit unnecessary but I might have a look at the implementation to see why that is.

mario-galic: `Set[A]` is certainly a subtype of `Iterable[A]`

```
implicitly[Set[CohortSpec] <:< Iterable[CohortSpec]]  // ok
```

I think the issue here is variance where sets are [invariant](https://stackoverflow.com/a/6183115/5205022) in `A`

```
trait Set[A]
```

but the type constraints on [`filter`](https://github.com/zio/zio/blob/3950efa0d94c870ed3487c3b438f3dcd2b3d77b2/core/shared/src/main/scala/zio/ZIO.scala#L2650) require covariance `Collection[+Element] <: Iterable[Element]`

```
def filter[R, E, A, Collection[+Element] <: Iterable[Element]](
    as: Collection[A]
  )(...
```

so `cohortSpecs.toList` works because `List` are covariant

```
sealed abstract class List[+A]
```

kelvin-chappell: Interesting.  But I still don't understand why the argument type needed to change.

mario-galic: I think previously they would always return `List` as in `ZIO[R, E, List[B]]`, but now they have generalised to (almost) any collection `ZIO[R, E, Collection[B]]` where `Collection` is type parameter. Unfortunately it does not work for `Set` because of the variance issue. However looking at [foreach](https://github.com/zio/zio/pull/4037/commits/ef31bf2cc888d3e3e22d5f1985749c4beaa60fbb#diff-9cfb50d60ba29ffb4a6a15c933dbf2c3R2727) I can see they provided special cases for `Set`, so maybe we could open ZIO pull request to add similar specialisations for `filter`. For example in `ZIO.scala` add

```
  def filter[E, A](as: Set[A])(f: A => IO[E, Boolean]): IO[E, Set[A]] =
    ZIO.filter(as)(f)

  def filterPar[E, A](as: Set[A])(f: A => IO[E, Boolean]): IO[E, Set[A]] =
    ZIO.filterPar(as)(f)

  def filterNot[E, A](as: Set[A])(f: A => IO[E, Boolean]): IO[E, Set[A]] =
    ZIO.filterNot(as)(f)

  def filterNotPar[E, A](as: Set[A])(f: A => IO[E, Boolean]): IO[E, Set[A]] =
    ZIO.filterNotPar(as)(f)
``` 

Then we could remove the `.toList` call.

kelvin-chappell: That sounds like a good idea.  I'd prefer not to have to convert to list unnecessarily.

PR #165, opened 2020-09-04
Retry Zuora calls on failure

kelvin-chappell:
Previously only the call to [fetch invoice preview after updating a subscription](https://github.com/guardian/price-migration-engine/compare/kc-zuora-retry?expand=1#diff-7e7faab75b94ac20b1e6efd5d0f2b264L77-L79) had a retry attached to it.
Now [all Zuora calls are retried](https://github.com/guardian/price-migration-engine/compare/kc-zuora-retry?expand=1#diff-04e584920e7bd8528c5096efbec9ce3cR105-R118) if they fail.

The [retry](https://github.com/guardian/price-migration-engine/compare/kc-zuora-retry?expand=1#diff-04e584920e7bd8528c5096efbec9ce3cR110) uses an exponential backoff policy - retrying up to 5 times at extending intervals starting at 1 second.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,3 +1,3 @@
-version = 2.6.4
+version = 2.7.0
 
 maxColumn = 120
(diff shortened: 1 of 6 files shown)


mario-galic:
Is it always safe to retry PUT and POST? It seems at least update subscription is atomic so should be safe https://www.zuora.com/developer/api-reference/#operation/PUT_Subscription

kelvin-chappell:
@mario-galic That's a very good point.  The sub update is atomic but we could potentially get multiple updates.  Also applying the retry generically for all requests is dangerous if this code gets copied and pasted or extended in use.  Also my implementation is wrong because it will only retry if the call throws an exception.  

So I'm going to do this differently and apply the retry only on Get requests, and otherwise on particular requests downstream such as the Post request to fetch the invoice preview.

mario-galic (approved):

PR #167, opened 2020-10-07
Bump ZIO to v1.0.2

kelvin-chappell:
This update brings in https://github.com/zio/zio/pull/4188, which means we no longer need to convert sets to lists to filter them. 
It also changes some `foreach` to `foreach_`, which is more efficient as it doesn't bother to build up results.


ripecosta (approved):
👍

PR #194, opened 2021-03-12
Run Snyk as a GitHub Action

ripecosta:
This PR configures a GitHub Action to run Snyk. This action runs every day at 6 AM and on every push. If the branch it's running on is main then it will run `snyk monitor` (reports vulnerabilities to snyk.io), otherwise it will run `snyk test`. 
It also upgrades sbt to v1.4.0 so we can use `dependencyTree`.

Note: Because `dependencyTree` was missing, Snyk was unable to run correctly. For that reason I did not until this PR is merged to remove the Snyk config from TeamCity and GitHub.


kelvin-chappell (approved):
Thanks @ripecosta! 👍

PR #263, opened 2021-05-21
Upgrade AWS SDK to v2

kelvin-chappell:
This upgrades all the AWS services from v1 to v2.
These are the main references:
* https://docs.aws.amazon.com/sdk-for-java/latest/migration-guide/whats-different.html
* https://docs.aws.amazon.com/sdk-for-java/latest/migration-guide/aws-sdk-java-mg.pdf
* https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md

Doing this should remove some of the vulnerable transitive dependencies that v1 had, which will keep Snyk happy.

It has been tested on the Dev stage.

Along the way:
* Removal of some [redundant config](https://github.com/guardian/price-migration-engine/compare/kc-aws-v2?expand=1#diff-f6015ed051d21ff9066e0e0c7fdb8f559d3bc34421daf6dc57092341266c7fb2L9-L12)
* Updated [gitignore](https://github.com/guardian/price-migration-engine/compare/kc-aws-v2?expand=1#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R3)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 .idea
+.bsp
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,20 @@ lazy val lambda = (project in file("lambda"))
     riffRaffUploadArtifactBucket := Option("riffraff-artifact"),
     riffRaffUploadManifestBucket := Option("riffraff-builds"),
     riffRaffManifestProjectName := "MemSub::Subscriptions::Lambda::PriceMigrationEngine",
-    riffRaffArtifactResources += ((project.base / "cfn.yaml", "cfn/cfn.yaml"))
+    riffRaffArtifactResources += ((project.base / "cfn.yaml", "cfn/cfn.yaml")),
+    assembly / assemblyMergeStrategy := {
+      /*
+       * AWS SDK v2 includes a codegen-resources directory in each jar, with conflicting names.
+       * This appears to be for generating clients from HTTP services.
+       * So it's redundant in a binary artefact.
+       */
+      case PathList("codegen-resources", _*)                    => MergeStrategy.discard
+      case PathList("module-info.class")                        => MergeStrategy.discard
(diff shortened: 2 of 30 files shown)


ripecosta (commented):
👍

paulbrown1982:
I'm amazed how much code is needed to upgrade this. Is it possible to comment where you've done any opportunistic refactoring which was not required to accommodate the upgrade?

Also, his is a v v old article, but it explains that the v2 SDK has a mechanism to stop the HTTP client from doing classpath scanning: https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-0-developer-preview/

See paragraph: `... Customers who want to avoid potentially expensive classpath scanning can set the system property software.amazon.awssdk.http.service.impl to explicitly identify the implementation to use...`

Is that something we can test?

kelvin-chappell:
@paulbrown1982 

> I'm amazed how much code is needed to upgrade this. Is it possible to comment where you've done any opportunistic refactoring which was not required to accommodate the upgrade?

The only opportunistic refactoring is around the removal of the redundant config, which is quite minimal.  Most of the change in terms of LOC is test code.
The AWS services are largely encapsulated in a single wrapper object per service but there is some leakage of the Dynamo DB service in various places, which could be optimised a little.

> Also, his is a v v old article, but it explains that the v2 SDK has a mechanism to stop the HTTP client from doing classpath scanning: https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-0-developer-preview/
> 
> See paragraph: `... Customers who want to avoid potentially expensive classpath scanning can set the system property software.amazon.awssdk.http.service.impl to explicitly identify the implementation to use...`
> 
> Is that something we can test?

Thanks for this link.  I wasn't aware of the the HTTP client integration.  I'll certainly have a go at specifying explicitly the HTTP client to use.

PR #513, opened 2022-03-29
Bump ZIO to v2

kelvin-chappell:
This is following the documentation here:
https://zio.dev/next/howto/migrate/zio-2.x-migration-guide

There's no change in functionality or in the signatures of public methods.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/CohortHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/CohortHandler.scala
@@ -1,17 +1,17 @@
 package pricemigrationengine.handlers
 
-import java.io.{InputStream, OutputStream}
-
 import com.amazonaws.services.lambda.runtime.{Context, RequestStreamHandler}
 import pricemigrationengine.model._
 import pricemigrationengine.services._
 import ujson.Readable
 import upickle.default.{read, stream}
-import zio.{ExitCode, Runtime, ZEnv, ZIO}
+import zio.{Runtime, ZEnv, ZIO, ZIOAppArgs, ZIOAppDefault}
+
+import java.io.{InputStream, OutputStream}
 
 /** A CohortHandler can be run as the handler of an AWS lambda or as a standalone program.
   */
-trait CohortHandler extends zio.App with RequestStreamHandler {
+trait CohortHandler extends ZIOAppDefault with RequestStreamHandler {
 
   /** Makes implementation available in lambda or console context.
     *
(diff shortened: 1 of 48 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -37,8 +37,8 @@ object NotificationHandler extends CohortHandler {
       )
       count <-
         subscriptions
-          .mapM(sendNotification(brazeCampaignName))
-          .fold(0) { (sum, count) => sum + count }
+          .mapZIO(sendNotification(brazeCampaignName))
+          .runFold(0) { (sum, count) => sum + count }
```
jonflynng, line 41: Where have you found this change from? I can't see it in here: https://zio.dev/next/howto/migrate/zio-2.x-migration-guide/

lambda/src/main/scala/pricemigrationengine/handlers/CohortHandler.scala
```diff
@@ -32,33 +32,33 @@ trait CohortHandler extends zio.App with RequestStreamHandler {
...
-          .filterOrElse(_.exists(CohortSpec.isValid))(spec => ZIO.fail(InputFailure(s"Invalid cohort spec: $spec")))
+          .filterOrElseWith(_.exists(CohortSpec.isValid))(spec => ZIO.fail(InputFailure(s"Invalid cohort spec: $spec")))
       validSpec <- ZIO.fromOption(cohortSpec).orElseFail(InputFailure("No input"))
     } yield validSpec).tapBoth(
       e => Logging.error(e.toString),
       spec => Logging.info(s"Input: $spec")
     )
 
-  final def run(args: List[String]): ZIO[ZEnv, Nothing, ExitCode] =
+  override final def run: ZIO[ZEnv with ZIOAppArgs, Any, Any] =
```
jonflynng, line 44: why has this been changed to override?

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
kelvin-chappell: This was done by the scalafix step because `fold` has now been deprecated in favour of `runFold` for `ZStreams`.  I think this is to make differences between `ZIO` and `ZStream` ops clearer and to make `ZStream` ops more consistent.
https://zio.dev/next/datatypes/stream/zstream/#using-fold

lambda/src/main/scala/pricemigrationengine/handlers/CohortHandler.scala
kelvin-chappell: It was always overriding the `run` method in the trait.  I just thought it might be better to make that explicit so that the compiler can help us out.

project/plugins.sbt
```diff
@@ -1,3 +1,4 @@
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "1.2.0")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.6")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.34")
```
jonflynng, line 4: what is this for?

kelvin-chappell: It was used for the migration but I can get rid of it now.

jonflynng (approved):

PR #515, opened 2022-03-30
Set a unique build number

kelvin-chappell:
Now that GHA is generating builds, the numbering has restarted from 1.  This means that build numbers are duplicating TeamCity build numbers.

We can ensure build numbers are unique by starting build numbers where the TeamCity ones ended.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,11 @@ jobs:
 
       - name: Upload artefacts to Riffraff
         if: env.isPrFromForkedRepo == 'false'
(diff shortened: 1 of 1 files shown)


kelvin-chappell:
This won't work with recent releases of the sbt-riffraff-artifact plugin.

jonflynng:
[TASK: Create tests for all Newspaper Delivery rate plan variations](https://trello.com/c/zQLWyh8x/3029-task-create-tests-for-all-newspaper-delivery-rate-plan-variations)

PR #516, opened 2022-03-30
Rename Riffraff projects to avoid duplicate build numbers

kelvin-chappell:
Following instructions here:
https://github.com/guardian/sbt-riffraff-artifact#migrating-riff-raff-project-to-github-actions-from-teamcity


jonflynng (approved):
looks good. I see the docs mention `riffRaffPackageName ` which isn't here, is there a need to have it as well as the project name?

kelvin-chappell:
> I see the docs mention `riffRaffPackageName ` which isn't here, is there a need to have it as well as the project name?
I think it's for a different type of deployment so it isn't needed here.

PR #519, opened 2022-03-31
Newspaper Delivery tests

jonflynng:
Test cases have been added to prepare for price migration of newspaper delivery subscriptions. 

- Tests have all run successfully without the need to change any code. 
- Tests for each newspaper delivery rateplan has been included.
- Tests for pre-2020 acquired subscriptions which did not go through a price-rise have been added
- Tests for newly acquired subscriptions after the latest price-rise have been added
- Tests with percentage discounts have been added


kelvin-chappell (approved):
I'm amazed that there are no code changes needed but these tests look very thorough.
Perhaps we should just keep one copy of the product catalogue but can think about that another time.

I think this is complete for standard home delivery and we should probably do some tests for echo legacy in a separate PR.
🚀

jonflynng:
[TASK: Create tests for all Newspaper Delivery rate plan variations](https://trello.com/c/zQLWyh8x/3029-task-create-tests-for-all-newspaper-delivery-rate-plan-variations)

PR #542, opened 2022-04-13
Remove redundant tmpTableName field from cohort spec

kelvin-chappell:
The first cohort had a hardcoded table name, which is no longer required.  Now we generate the table name from the cohort name.


jonflynng (approved):
looks good.

PR #548, opened 2022-04-21
add cohort name to logging

jonflynng:
As we are running the state machines to process multiple cohorts, we are adding extra logging information so we are able to distinguish between cohort's when querying via log insights in Cloudwatch.

Logging the message in JSON format allows us to query keys within the JSON in Log Insights. An example @message log field would look like so:

`{"CohortName": "Vouchers2022", "some_key": "INFO: Successfully updated Price_Rise__c object: a2d5I000000YYoOQAW"}`

An example query would be: 

```
fields @timestamp, @message, CohortName
| filter CohortName = "Vouchers2022"
| sort @timestamp desc
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,6 +39,8 @@ lazy val lambda = (project in file("lambda"))
       awsSQS,
       awsStateMachine,
       http,
+      circeGeneric,
+      circeParser,
       commonsCsv,
       slf4jNop % Runtime,
       munit % Test,
--- a/build.sbt
+++ b/build.sbt
@@ -39,8 +39,6 @@ lazy val lambda = (project in file("lambda"))
       awsSQS,
       awsStateMachine,
       http,
-      circeGeneric,
-      circeParser,
       commonsCsv,
       slf4jNop % Runtime,
       munit % Test,
--- a/lambda/src/main/scala/pricemigrationengine/services/LambdaLogging.scala
+++ b/lambda/src/main/scala/pricemigrationengine/services/LambdaLogging.scala
@@ -2,15 +2,28 @@ package pricemigrationengine.services
 
 import com.amazonaws.services.lambda.runtime.{Context, LambdaLogger}
 import zio.{UIO, ULayer, ZIO, ZLayer}
+import io.circe.generic.auto._
+import io.circe.syntax._
 
 object LambdaLogging {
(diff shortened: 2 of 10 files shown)


build.sbt
```diff
@@ -39,6 +39,8 @@ lazy val lambda = (project in file("lambda"))
       awsSQS,
       awsStateMachine,
       http,
+      circeGeneric,
+      circeParser,
```
kelvin-chappell, line 43: We're already using [upickle](https://com-lihaoyi.github.io/upickle/) as a json-handling library.

lambda/src/main/scala/pricemigrationengine/services/LambdaLogging.scala
```diff
@@ -2,15 +2,28 @@ package pricemigrationengine.services
...
+  def impl(context: Context, cohortName: String): ULayer[Logging] =
     ZLayer.succeed(
       new Logging.Service {
         val logger: LambdaLogger = context.getLogger
-        def info(s: String): UIO[Unit] = ZIO.succeed(logger.log(s"INFO: $s"))
-        def error(s: String): UIO[Unit] = ZIO.succeed(logger.log(s"ERROR: $s"))
+        def info(s: String): UIO[Unit] =
+          ZIO.succeed(logger.log(InfoMessage(cohortName, s).asJson.dropNullValues.noSpaces))
+        def error(s: String): UIO[Unit] =
+          ZIO.succeed(logger.log(ErrorMessage(cohortName, s).asJson.dropNullValues.noSpaces))
```
kelvin-chappell, line 26: There shouldn't be any nulls.
`write(...)` should work with upickle.

kelvin-chappell (commented):
I like the approach of constructing the LambdaLogger layer with the cohort name baked into it.  But I think you'll have to somehow overwrite the initial logging implementation at the point where the cohort spec is validated and then the cohort name is available.

kelvin-chappell (commented):
Great solution!  If you just get rid of circe I'll approve it

jonflynng:
[Price-Rise: Add the cohort name to logs](https://trello.com/c/rF9J14qy/3099-price-rise-add-the-cohort-name-to-logs)

kelvin-chappell (approved):
👍

PR #555, opened 2022-04-29
Migrate echo-legacy home delivery subscriptions to new rateplans

jonflynng:
### Objective
There are around 700 newspaper delivery subscriptions which are using the old echo-legacy rateplan, i.e.: the customer chooses the days they want the paper delivered and is then charged for those days only. Most of these subscriptions map easily to some of our newer rateplans like Weekend, Everyday, Sixday, etc...

**To note:**

- There are 59 subscriptions which do not map to these new rateplans. These will not be included in the cohort, as Rob has said he will look at handling these separately.
- These subscriptions are either monthly or quarterly billed.
- [A spreadsheet with all subscriptions and the new rateplans they need to be matched with](https://docs.google.com/spreadsheets/d/1fQ9MGdQI0L68fJrHTSkU3uNh61Oyy4bT4ytcYrvhCKo/edit#gid=0)

### What has changed
We have had to add additional logic to match these subscriptions with their new rateplans. This has been included in a separate file; EchoLegacy.scala. This logic has then be used throughout the Estimation and Amendment lambdas.

 The added code is specific to Echo-Legacy plans and won't/can't be used anywhere else in the future.

### Testing
Unit tests have been added to AmendmentTest.scala and ZuoraSubscriptionUpdateTest.scala files, which support the Estimation and Amendment lambdas.


lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -104,31 +106,97 @@ object AmendmentData {
...
+          case 7 => getNewPlan("Everyday", chargedDays)
+          case 6 if chargedDays.filter(_.name == "Sunday").isEmpty =>
+            getNewPlan("Sixday", chargedDays)
+          case 2 if chargedDays.filter(plan => plan.name == "Saturday" || plan.name == "Sunday").length == 2 =>
+            getNewPlan("Weekend", chargedDays)
+          case 1 =>
+            chargedDays.head.name match {
+              case "Saturday" => getNewPlan("Saturday", chargedDays)
+              case "Sunday"   => getNewPlan("Sunday", chargedDays)
+              // is this necessary or will it default to the case below at line 154 if there is no match??
```
jonflynng, line 155: @kelvin-chappell

kelvin-chappell (commented):
Looks good!  I wonder if it's possible to separate out the Echo-legacy specific from the generic AmendmentData as it's a one-off and won't be used in future rises?  If not, could just remove after all amendments have gone through.

jonflynng:
[Price Rise - Migrate echo-legacy delivery subs to newer rateplans](https://trello.com/c/AyzVDmIG/3092-price-rise-migrate-echo-legacy-delivery-subs-to-newer-rateplans)

kelvin-chappell (approved):
Excellent! 🚀

PR #561, opened 2022-05-06
Upgrade ZIO v2 to RC3

kelvin-chappell:
The main change here is that `ZManaged` has been replaced by `Scope`.
This is a more straightforward way of managing opening and closing of resources.
Just have to be aware that the lifecycle of the use of a resource has to be inside a `ZIO.scoped` block.

See:

* https://github.com/zio/zio/releases/tag/v2.0.0-RC3
* https://zio.dev/next/howto/migrate/zio-2.x-migration-guide/#migration-from-zmanaged-to-scope


jonflynng (approved):
looks good. I found this to be a good explanation and reference for the future: https://www.youtube.com/watch?v=E1D7HiUKj_k

PR #565, opened 2022-05-10
Documentation Updates

jonflynng:


lambda/README.md
```diff
@@ -4,176 +4,201 @@
 
 ### Code structure
```
jonflynng, line 5: intelliJ re-formatted this file so sorry about the large diff here

cohort-setup.md
```diff
@@ -1,10 +1,12 @@
 # How to set up a new cohort
 
-These are the steps required to set up a new cohort of subscriptions due to have their price increased.
+Previously, cohorts have been separated by product type, i.e.: Newspaper Vouchers and Newspaper Delivery.
```
kelvin-chappell, line 3: That was just for convenience.  Cohorts don't actually need to be by product type.
```suggestion
Previously, cohorts have been separated by product type, i.e.: Newspaper Vouchers and Newspaper Delivery.  The cohort names reflected this, eg. 'Vouchers2022'.  This was because the engine was being developed to support new product types while it was in use.  But cohorts don't need to be product-specific and the whole lot could be put in a single cohort.  How the cohorts are broken down depends on which subscriptions are ready for migration, either for business or technical reasons.
```

cohort-setup.md
```diff
@@ -19,10 +21,47 @@ Format is `yyyy-mm-dd`.
...
 Add two CSV files to the folder:
 * **salesforce-subscription-id-report.csv**: A file holding line-separated subscription numbers to be **included** in the 
-price rise, which will typically be the result of a Salesforce report.
+price rise, which will typically be the result of a Salesforce report. More info in [ImportSubscriptionId.MD](./ImportSubscriptionId.MD) on this. 
 * **excluded-subscription-ids.csv**: A file holding line-separated subscription numbers to be **excluded** from the price 
 rise.
+
+## 3. Run the state machine
+
+Navigate to `price-migration-engine-cohort-steps-PROD` from the Step Functions dashboard. Click on Start Execution and configure the input JSON. Look at previous executions for examples of what the input JSON should look like. E.g.: `{"cohortSpec":{"cohortName":"HomeDelivery2022","brazeCampaignName":"SV_PrintPriceRise_2022","importStartDate":"2022-04-28","earliestPriceMigrationStartDate":"2022-06-29"}}`
```
kelvin-chappell, line 34: Might be easier to run the kick-off lambda [price-migration-lambda-PROD](https://eu-west-1.console.aws.amazon.com/lambda/home?region=eu-west-1#/functions/price-migration-lambda-PROD).  
That would have the advantage of testing if the cohort spec entry is correct as well.

cohort-setup.md
jonflynng: Yep, thanks for making it clearer

kelvin-chappell (approved):
👍

PR #572, opened 2022-05-16
Check scalafmt in PRs from forked repos

kelvin-chappell:


jonflynng (approved):
Great, will stop my builds always breaking :)

PR #577, opened 2022-05-18
Remove redundant service aliases

kelvin-chappell:
The essence of this change is the deletion of these [redundant service aliases](https://github.com/guardian/price-migration-engine/compare/kc-aliases?expand=1#diff-a5c100fbc1dee8b34f2d8dd3ac83b5f2a3359f0f300b7219c7118f19a8debc35L1-L15).  
They were required by ZIO v1 but no longer in ZIO v2.
The rest of the PR is adjusting the types so that they don't rely on the deleted types.

See https://zio.dev/next/howto/migrate/zio-2.x-migration-guide/#has


jonflynng (approved):
nice, definitely looks cleaner

PR #578, opened 2022-05-19
Quarterly fix for estimating echo legacy

jonflynng:
**Problem**

29 subscriptions failed estimation when running through the EchoLegacy cohort. We've identified a bug with estimating quarterly plans:

[Take an EchoLegacy -> Saturday quarterly plan for example] (https://github.com/guardian/price-migration-engine/blob/main/lambda/src/test/resources/NewspaperDelivery/EchoLegacy/SaturdayQuarterly/InvoicePreview.json)

The saturday rateplancharge is being billed quarterly here. However, the rest of the days are being billed monthly. Here's a sixday plan which shows the rateplan charges in the Zuora UI:

Screenshot 2022-05-19 at 10 29 08

This means invoices are being generated every single month still, just with a charge of 0 until the quarter comes around.

The price-rise engine was then looking at the next invoice after the earliest start date, which will have not contained all necessary rateplancharges to estimate correctly.

**Solution**

the nextServiceStartDateEchoLegacy function groups the invoices by date and then ensures there are 7 rateplancharges, for the given date. The groups with less than 7 charges will be discarded.

**Important Note**

This PR will have to be reverted after the estimations have gone through.


lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -33,11 +33,16 @@ object AmendmentData {
       subscription: ZuoraSubscription,
       invoiceList: ZuoraInvoiceList,
       earliestStartDate: LocalDate
-  ): Either[AmendmentDataFailure, AmendmentData] =
+  ): Either[AmendmentDataFailure, AmendmentData] = {
+    val isEchoLegacy = subscription.ratePlans.filter(_.ratePlanName == "Echo-Legacy").length >= 1
```
kelvin-chappell, line 37:
```suggestion
    val isEchoLegacy = subscription.ratePlans.filter(_.ratePlanName == "Echo-Legacy").nonEmpty
```

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -52,6 +57,26 @@ object AmendmentData {
       .headOption
       .toRight(AmendmentDataFailure(s"Cannot determine next billing date on or after $onOrAfter from $invoiceList"))
 
+  def nextServiceStartDateEchoLegacy(
```
kelvin-chappell, line 60: Perhaps a comment to explain why this is necessary

kelvin-chappell (approved):
👍

PR #584, opened 2022-05-20
Add new state for subs where price isn't rising

kelvin-chappell:
Previously, subs where the new price wasn't greater than the old price were considered to be estimation failures.  This meant they kept coming up when we analysing failure cases.  But we didn't want to remove them from the cohort as that would then raise the question why they had been excluded.

To solve the problem, in this change we no longer consider price falls or price constant as failures.  
Instead they are given a new state in the cohort table, 'NoPriceIncrease', which isn't processed any further.
This will allow us to analyse them separately if necessary.


jonflynng (approved):
looks good!

PR #585, opened 2022-05-23
Add new NoPriceIncrease state to set of all

kelvin-chappell:
We're seeing an alarm go off because [this code](https://github.com/guardian/price-migration-engine/blob/main/lambda/src/main/scala/pricemigrationengine/model/dynamodb/Conversions.scala#L60-L62) does a check for invalid subscription processing states before exporting data to the lake.
Error is:
> DynamoDBZIOError(The 'processingStage' contained an invalid CohortTableFilter 'NoPriceIncrease')


jonflynng (approved):
Can you just add in the comment stating why this was triggering the alarm.

PR #586, opened 2022-05-27
Run CI on every push

kelvin-chappell:
This will give some confidence before raising a PR.


kelvin-chappell:
On second thoughts, don't like this idea because it means doing multiple redundant CI runs in some cases.

PR #587, opened 2022-05-27
Flatten cohort stream structure

kelvin-chappell:


kelvin-chappell:
Will make separate PRs for `CohortTable.fetch` and `CohortTable.fetchAll`.

PR #590, opened 2022-05-27
Eliminate LiveLayer.exportConfig

kelvin-chappell:
This is the first step in eliminating the LiveLayer object as there's no longer any need for complex layer building in ZIO v2.

See https://zio.dev/next/howto/migrate/zio-2.x-migration-guide/#building-the-dependency-graph


jonflynng (approved):
nice, a big improvement in readability

PR #662, opened 2022-08-08
Remove echo legacy code

jonflynng:
Removes the code we used to migrate Echo Legacy rateplans to the newer Newspaper Delivery ones. Deleting this as it will never be used again and don't want to clutter up the codebase. 

 [The original Echo Legacy PR](https://github.com/guardian/price-migration-engine/pull/555)


shtukas (approved):
✂️

PR #677, opened 2022-08-22
Add currency code as prefix of the payment amount

shtukas:
Using the currency ISO code to symbol conversion found here ( https://github.com/guardian/support-frontend/blob/fc932544ff7730b9cec71e562a5f6ca12a08694c/support-internationalisation/src/main/scala/com/gu/i18n/Currency.scala ), we update the `payment_amount` field to be prefixed with the corresponding currency symbol.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -5,6 +5,7 @@ import pricemigrationengine.model._
 import pricemigrationengine.model.membershipworkflow._
 import pricemigrationengine.services._
 import zio.{Clock, ZIO}
+import com.gu.i18n
 
 object NotificationHandler extends CohortHandler {
 
@@ -52,6 +53,10 @@ object NotificationHandler extends CohortHandler {
     }
   }
 
+  def currencyISOtoSymbol(iso: String): ZIO[Any, Nothing, String] = {
+    ZIO.succeed(i18n.Currency.fromString(iso: String).map(_.identifier).getOrElse(""))
+  }
+
   def sendNotification(
       brazeCampaignName: String,
       cohortItem: CohortItem,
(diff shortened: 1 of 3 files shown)


shtukas:
Note that considering how `currencyISOtoSymbol` is currently written, if the currency given (which should be a ISO value) is not supported by our mapping (this should technically not happen), then the payment_amount will not carry any prefix.

jonflynng (approved):

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -52,6 +53,10 @@ object NotificationHandler extends CohortHandler {
     }
   }
 
+  def currencyISOtoSymbol(iso: String): ZIO[Any, Nothing, String] = {
+    ZIO.succeed(i18n.Currency.fromString(iso: String).map(_.identifier).getOrElse(""))
+  }
```
kelvin-chappell, line 58: If this can't fail, it doesn't need to be wrapped in a `ZIO`:
```
private def currencyISOtoSymbol(iso: String): String = 
    i18n.Currency.fromString(iso: String).map(_.identifier).getOrElse("")
```

but arguably it should fail or put the sub into a different state if there's no corresponding symbol (eg if some new currency emerges.)

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -72,6 +77,9 @@ object NotificationHandler extends CohortHandler {
       startDate <- requiredField(cohortItem.startDate.map(_.toString()), "CohortItem.startDate")
       billingPeriod <- requiredField(cohortItem.billingPeriod, "CohortItem.billingPeriod")
       paymentFrequency <- paymentFrequency(billingPeriod)
+      currencyISOCode <- requiredField(cohortItem.currency, "CohortItem.currency")
+      currencySymbol <- currencyISOtoSymbol(currencyISOCode)
```
kelvin-chappell, line 81: If `currencyISOtoSymbol` as above:
```
currencySymbol = currencyISOtoSymbol(currencyISOCode)
```

PR #686, opened 2022-08-30
start date in user friendly format

shtukas:
At the moment the start date is read from the database in standard ISO YYYY-MM-DD format and echoed as such in the letter. This can cause problems to users unfamiliar with that format (for instance US readers).

This change  introduce the `startDateConversion` function which takes such as string, for instance `2022-08-30` and returns the more friendly `30 August 2022`. The functions uses two helpers (instead of being written in one single self contained function) to allow the start date test data to carry on being defined as a LocalDate.

### DIFF
--- a/lambda/src/test/scala/pricemigrationengine/handlers/NotificationHandlerTest.scala
+++ b/lambda/src/test/scala/pricemigrationengine/handlers/NotificationHandlerTest.scala
@@ -1,6 +1,6 @@
 package pricemigrationengine.handlers
 
-import pricemigrationengine.TestLogging
+import pricemigrationengine.{TestLogging, handlers}
 import pricemigrationengine.model.CohortTableFilter._
 import pricemigrationengine.model._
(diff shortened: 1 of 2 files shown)


lambda/src/test/scala/pricemigrationengine/handlers/NotificationHandlerTest.scala
```diff
@@ -19,6 +19,7 @@ import scala.collection.mutable.ArrayBuffer
 class NotificationHandlerTest extends munit.FunSuite {
   private val expectedSubscriptionName = "Sub-0001"
   private val expectedStartDate = LocalDate.of(2020, 1, 1)
+  private val expectedStartDateUserFriendlyFormat = "01 January 2020"
```
paulbrown1982, line 22: I think we should drop the leading 0, so it should be "1 January 2020".

jonflynng (approved):

PR #701, opened 2022-09-14
Use customerAcceptanceDate in spreadEarliestStartDate

shtukas:
This PR is a follow up of https://github.com/guardian/price-migration-engine/pull/698 . We incorporates the suggestion of using ~contractAcceptanceDate~ customerAcceptanceDate instead of `termStartDate` for in in spreadEarliestStartDate.


jonflynng (approved):
I'm pretty sure this is correct to use the `customerAcceptanceDate` over the `contractEffectiveDate` or `termStartDate`.

 The contractEffectiveDate is the date the Zuora subscription/amendment is created, and the customerAcceptanceDate is when it actually becomes active.

 Couple examples:

 1. This sub (A-S01255801) has just gone through a price-rise. The contractEffectiveDate seems to be the date the engine added the amendment to Zuora, but there is still a month until the new ratePlan becomes active.
2. When we create a new digital subscription, we add 14 days to the customerAcceptanceDate to signify a free trial.

PR #708, opened 2022-09-20
All monthly subscriptions less than one year old to be risen on their 13th month

jonflynng:
This is to fix the generation of start dates of subscriptions less than one year old introduced in [this PR](https://github.com/guardian/price-migration-engine/pull/698)

With the old code, when running a test cohort of GW subscriptions, we get the next billing date for a subscription and add between 12 and 15 months onto it to obtain the startDate. This means all start dates are around the start of 2024 (this is wrong):

Screenshot 2022-09-20 at 16 15 15

To be clear, any subscription that is less than one year old, must be price-risen as soon as their first year is over, i.e.: on the 13th month.    @paulbrown1982 mentioned this before but now only that I have run some subscriptions through the estimation process I've been able to visualise what the resulting start dates would be.

Tests have been provided for this.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
@@ -95,7 +95,12 @@ object EstimationHandler extends CohortHandler {
         yearAgo <- Clock.localDateTime.map(_.toLocalDate.minusYears(1))
         randomFactor <-
           if (subscription.customerAcceptanceDate.isBefore(yearAgo)) Random.nextIntBetween(0, 3)
-          else ZIO.succeed(relu(ChronoUnit.MONTHS.between(earliestStartDate, subscription.customerAcceptanceDate.plusMonths(13)).toInt))
+          else
+            ZIO.succeed(
+              relu(
+                ChronoUnit.MONTHS.between(earliestStartDate, subscription.customerAcceptanceDate.plusMonths(13)).toInt
+              )
+            )
         actualEarliestStartDate = earliestStartDate.plusMonths(randomFactor)
       } yield actualEarliestStartDate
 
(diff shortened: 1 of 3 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -80,18 +81,26 @@ object EstimationHandler extends CohortHandler {
...
    * Earliest start date spread out over 3 months.
    */
-  private[handlers] def spreadEarliestStartDate(
+  def spreadEarliestStartDate(
       subscription: ZuoraSubscription,
       invoicePreview: ZuoraInvoiceList,
       earliestStartDate: LocalDate
   ): IO[ConfigFailure, LocalDate] = {
 
+    def relu(number: Int): Int =
```
jonflynng, line 90: the `.between` function will return a negative integer if the `customerAcceptanceDate.plusMonths(13)` is before the `earliestStartDate`. This will ensure those subscriptions do not get a start date earlier than the earliestStartDate.

lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -80,18 +81,26 @@ object EstimationHandler extends CohortHandler {
...
         yearAgo <- Clock.localDateTime.map(_.toLocalDate.minusYears(1))
         randomFactor <-
           if (subscription.customerAcceptanceDate.isBefore(yearAgo)) Random.nextIntBetween(0, 3)
-          else Random.nextIntBetween(12, 15)
+          else
+            ZIO.succeed(
+              relu(
+                ChronoUnit.MONTHS.between(earliestStartDate, subscription.customerAcceptanceDate.plusMonths(13)).toInt
+              )
+            )
```
kelvin-chappell, line 103: Maybe move this block into its own function with some commentary as it's not immediately obvious what the intention is here

kelvin-chappell (approved):
Logic makes sense 👍

PR #727, opened 2022-10-09
Introduce ChargeCap

shtukas:
Out of consideration for our users, we have the overall policy to not increase subscriptions charges by more than a certain amount relatively to the old price. At the time these lines are written, we cap it to 20% increase. (This means that if the current/old price is 100 and the new price computed after applying a ew rate plan if 130, then we limit to 120.)

To encode this cap, we introduce `ChargeCap`, which essentially carries a capped price. This object is passed, optionally, to any function that would need to know about it.

The Amendment Handler uses `ChargeCap` twice. The first time is to compute the correct amendment, the second time is to check that the price after the amendment is within parameters. 

The Estimation Handler is slightly more subtle. we do not at first know the old price (which will need to be computed as part of the estimation) In this case, we pass down a `ChargeCap.builderFromMultiplier` which is going to compute the relevant `ChargeCap` at first opportunity and will use it to compute the estimated price.

Something else to notice is that in the case of a non capped price increase, both the EstimationHandler and the AmendmentHandler do exacting the same as before. 

### Follow up 

Once this goes live, we will 
1. move the subscriptions currently stuck in "CappedPriceIncrease" state to "ReadyForEstimation".
2. remove the code that put them into "CappedPriceIncrease", in a subsequent PR, since we will not need it anymore.


shtukas:
I should probably write some tests to check that, in the case of a capped in increase, the AmendmentHandler is going to modify the subscription in a way that is consistent with the capped computation that is done by the EstimationHandler.

kelvin-chappell:
> I should probably write some tests to check that, in the case of a capped in increase, the AmendmentHandler is going to modify the subscription in a way that is consistent with the capped computation that is done by the EstimationHandler.

Might be a good idea in general to throw an exception in some way (either by stopping the engine or putting the sub into a different state, but raising an alarm in either case) when an estimated result disagrees with the actual result so that it can be investigated (and potentially corrected/reversed.)

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -41,7 +42,18 @@ object AmendmentHandler extends CohortHandler {
   private def doAmendment(
       catalogue: ZuoraProductCatalogue,
       item: CohortItem
-  ): ZIO[Zuora, Failure, SuccessfulAmendmentResult] =
+  ): ZIO[Zuora, Failure, SuccessfulAmendmentResult] = {
+
+    def subscriptionUpdatePriceOverride(oldPrice: BigDecimal, newPrice: BigDecimal): Option[BigDecimal] = {
```
kelvin-chappell, line 47:
```suggestion
    def subscriptionUpdateChargeOverride(oldPrice: BigDecimal, newPrice: BigDecimal): Option[BigDecimal] = {
```

kelvin-chappell (commented):
General principle looks good.
I think it might be less confusing to refer to `chargeOverride` throughout rather than `priceOverride` so it matches the Zuora model.
Also it looks like the [CappedPriceIncrease](https://github.com/guardian/price-migration-engine/blob/main/lambda/src/main/scala/pricemigrationengine/model/CohortTableFilter.scala#L50) type can go now.

And, more generally, with more special cases the code is beginning to be quite hard to follow for anyone unfamiliar with it.  I wonder if some high-level diagrams or commentary might help (but not in this PR).

lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -18,6 +19,7 @@ object EstimationHandler extends CohortHandler {
 
   // TODO: move to config
   private val batchSize = 150
+  private val priceCappingMultiplier = 1.2
```
kelvin-chappell, line 22: Should move this to a common location to avoid it changing between estimation and amendment.

kelvin-chappell (approved):
Just a comment about sharing the multiplier between `EstimationHandler` and `AmendmentHandler` but otherwise good 👍

lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
shtukas: Yes. I intend to move both in a PR shortly.

PR #744, opened 2022-11-22
Charge cap fix 

jonflynng:
Charge overrides with price caps were not being applied due to [this line](https://github.com/guardian/price-migration-engine/blob/b99f0601349e37d4114e21a075b4bdf94c87f5b3/lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscriptionUpdate.scala#L266) in the code. This has been refactored and we've introduced a boolean that tells us whether the current subscription has been price capped, if it has, we create the charge override.

 This affected 4 subscriptions. We manually removed any amendments created by the engine in the Zuora UI. The engine will pick them up again when it runs.


shtukas (approved):

shtukas:
Amazing, that was a fun pairing 🙂

PR #746, opened 2022-11-23
Better AmendmentDataFailure error message

shtukas:
Add the subscription number to the AmendmentDataFailure's message when no rate plans could be found, to help with investigations.


kelvin-chappell (approved):
👍

PR #747, opened 2022-11-23
Introduce MinNotificationLeadTime

shtukas:
In this PR we address the problem of the possibility of (letter) notifications being sent too close to a start date.

We have the convention to allow 45 days between Salesforce update and the actual start date of a price increase, this is to allow for the letters to be sent on time. 

**Problem**

The aim was to make the `earliestPriceMigrationStartDate` less brittle because if set wrong, we can potentially change thousands of subscriptions with no lead time. There's a couple of other cases which could go wrong:

- It can also happen though that some problem occurs, a subscription needs to be looked at, the price migration engine has a bug that needs fixing, etc, and in those cases the notification steps happens too late (meaning too close to the "start date"). 
- It may also be that someone wishes to set failed records back to `ReadyForEstimation` but forget to push back the `earliestPriceMigrationStartDate` in the cohort spec/parent table. 

To solve this problem we introduce a check in the notification handler.


lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -29,9 +34,23 @@ object AmendmentHandler extends CohortHandler {
...
+          // is not recorded as a failure of `amend` to allow the processing to continue.
           val result = CancelledAmendmentResult(item.subscriptionName)
           CohortTable.update(CohortItem.fromCancelledAmendmentResult(result)).as(result)
+        }
+        case _: StartDateNeedsToBeUpdatedFailure => {
+          // `StartDateNeedsToBeUpdatedFailure` is returned when we detect that the start date of an item is too close
+          // to the current date. This happens when, for instance, too long has passed between the creation of the
+          // cohort item and the actual Amendment. The side effect here is to update the item start date and to send
+          // it back to state `ReadyForEstimation`
+          val startDateNew = item.startDate.getOrElse(LocalDate.now()).plusDays(startDateShiftInCaseOfShortLeadTime)
```
kelvin-chappell, line 50: Best to avoid relative times as it's very difficult to test.  Use `zio.Clock` instead to get current time.

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -44,8 +63,16 @@ object AmendmentHandler extends CohortHandler {
       item: CohortItem
   ): ZIO[Zuora, Failure, SuccessfulAmendmentResult] = {
 
+    def checkStartDate(startDate: LocalDate, subscriptionName: String): ZIO[Any, Failure, Unit] = {
+      if (LocalDate.now().plusDays(startDateLeadTimeRequirement).isAfter(startDate)) ZIO.succeed(())
```
kelvin-chappell, line 67: Best to avoid relative times as it's very difficult to test. Use zio.Clock instead to get current time.

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -3,15 +3,20 @@ package pricemigrationengine.handlers
...
+import pricemigrationengine.model.CohortTableFilter.ReadyForEstimation
 import zio.{Clock, ZIO}
 
+import java.time.{Instant, LocalDate}
+
 /** Carries out price-rise amendments in Zuora.
   */
 object AmendmentHandler extends CohortHandler {
 
   // TODO: move to config
```
kelvin-chappell, line 15: Perhaps this should be removed if these vals are staying here

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -29,9 +34,23 @@ object AmendmentHandler extends CohortHandler {
   ): ZIO[CohortTable with Zuora, Failure, AmendmentResult] =
     doAmendment(catalogue, item).foldZIO(
       failure = {
-        case _: CancelledSubscriptionFailure =>
+        case _: CancelledSubscriptionFailure => {
+          // `CancelledSubscriptionFailure` happens when the subscription was cancelled in Zuora
+          // in which case we simply update the processing state for this item in the database
+          // Although it was given to us as a failure of `doAmendement`, the only effect if the database update, it
+          // is not recorded as a failure of `amend` to allow the processing to continue.
```
kelvin-chappell, line 41: Just picking holes here!
```suggestion
          // Although it was given to us as a failure of `doAmendment`, the only effect of the database update, if it
          // is not recorded as a failure of `amend`, is to allow the processing to continue.
```

kelvin-chappell (commented):
Looks good but relative dates can cause a lot of problems and you might want to add a test to prove it works.

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
jonflynng: Use `Clock.currentDateTime.map(_.toLocalDate)` instead of `LocalDate.now()`. Then we can use zio's [test clock](https://github.com/guardian/price-migration-engine/blob/0d252dc032fcfce1933a0ce7dd10da181f7b4ff5/lambda/src/test/scala/pricemigrationengine/handlers/EstimationHandlerTest.scala#L18) to set the time in tests.

jonflynng: the startDate can't be calculated here. We need to let the estimation lambda generate it, so we just need to set it back to ReadyForEstimation and then change logic in the estimation handler so the earliestStartDate is 45 days forward (in this case Jess said we're only doing a 30 day lead time).

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
jonflynng: In theory this should be fine, we send notifications, and do the amendments exactly 49 days in advance or whatever the lead time is. But yeah, Kelvin's probably right, if we have any failures with the Notification lambda then this logic would fail. There would probably be a lot of edge cases.

Having a think, all we need to be sure of is that the `startDate` is 49 days after the `whenNotificationSent`. So let's just compare those two fields.

@shtukas

jonflynng (changes requested):
In theory this should be fine, we send notifications, and do the amendments exactly 49 days in advance or whatever the lead time is. But yeah, Kelvin's probably right, if the engine fails for a few days in a row then this logic would fail. There would probably be a lot of edge cases.

Having a think, all we need to be sure of is that the `startDate` is 49 days after the `whenNotificationSent` date. So let's just compare those two fields. If `startDate` is within the lead time, then we could create a new field in the row called `newEarliestStartDate` or something like that. In the estimation lambda, we'd use that, and we'd default to using the `earliestPriceMigrationStartDate` from the Cohort Spec/parent table.

shtukas:
@jonflynng @kelvin-chappell 

Thank you for the comments and suggestions on the previous version of this. 

Talking with Jon we came up with the idea of an improvement to the engine that could be worth while to implement in the future, but today, we are simply addressing a pressing issue, and for this I changed the scope and purpose of that PR (and changed the code accordingly). See new PR description :)

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -14,10 +14,15 @@ object NotificationHandler extends CohortHandler {
...
-
   val Cancelled_Status = "Cancelled"
 
+  // We are starting the notification process for any item whose start date is less than 49 (NotificationLeadTimeDays)
+  // days away, but because it can happen, we also need to react to items which for *any* reason have
+  // not been processed on time (meaning early enough) and for which the previously computed startDate is
+  // too close (less than NotificationPaddingTimeDays away) from now and would not give enough time to the letters.
+
   private val NotificationLeadTimeDays = 49
+  private val NotificationPaddingTimeDays = 35
```
kelvin-chappell, line 25: Perhaps these should be called something more like `StandardNotificationLeadTime` and  `minNotificationLeadTime` and given a duration value.  I don't know if that's any clearer but padding seems difficult to understand to me.

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -32,27 +37,57 @@ object NotificationHandler extends CohortHandler {
...
-      for {
-        _ <- Logging.error(
-          s"Subscription ${cohortItem.subscriptionName}: Failed to send price rise notification: $failure"
+
+    // We are starting with a simple check. That the item's startDate is at least NotificationPaddingTimeDays days away
+    // from the current day. This will avoid headaches caused by letters not being sent early enough relatively to
+    // previously computed start dats, and will detect any such problem when they happen and that
+    // before the letters are sent.
+
+    if (itemHasEnoughNotificationPadding(LocalDate.now(), cohortItem)) {
```
kelvin-chappell, line 61: Should avoid relative dates

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -32,27 +37,57 @@ object NotificationHandler extends CohortHandler {
...
+            s"Subscription ${cohortItem.subscriptionName}: Failed to send price rise notification: $failure"
+          )
+        } yield Unsuccessful
+      }
+
+    } else {
+
+      ZIO.fail(
+        NotificationNotEnoughLeadTimeFailure(
+          s"The start date of item ${cohortItem.subscriptionName} (startDate: ${cohortItem.startDate}) is too close to today ${LocalDate.now()}"
```
kelvin-chappell, line 87: Relative date here too

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
jonflynng: @kelvin-chappell  by relative dates do you mean avoiding using `LocalDate.now()` and using Zio's clock instead or avoiding relative dates altogether?

 @shtukas   even though using `LocalDate.now()` works well for this function and is easily testable, it's best to use Zio clock incase we do integration tests

kelvin-chappell: Yes sorry I don't mean avoid relative dates altogether.  I just mean use zio.Clock to get them lazily instead.

shtukas: Corrected. Thanks!

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
shtukas: 👍

lambda/src/test/scala/pricemigrationengine/handlers/NotificationHandlerTest.scala
```diff
@@ -398,4 +399,16 @@ class NotificationHandlerTest extends munit.FunSuite {
     )
     assertEquals(sentMessages.size, 0)
   }
+
+  test("itemHasEnoughNotificationPadding behaves correctly") {
```
kelvin-chappell, line 403:
```suggestion
  test("thereIsEnoughNotificationLeadTime behaves correctly") {
```

kelvin-chappell (approved):
👍

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -14,45 +14,76 @@ object NotificationHandler extends CohortHandler {
...
+        .fetch(SalesforcePriceRiceCreationComplete, Some(today.plusDays(StandardNotificationLeadTime)))
+        .mapZIO(item => sendNotification(brazeCampaignName)(item, today))
         .runFold(0) { (sum, count) => sum + count }
       _ <- Logging.info(s"Successfully sent $count price rise notifications")
     } yield HandlerOutput(isComplete = true)
   }
 
+  def thereIsEnoughNotificationLeadTime(today: LocalDate, cohortItem: CohortItem): Boolean = {
+    // To help with backward compatibility with existing tests, we apply this condition from 1st Dec 2022.
+    if (today.isBefore(LocalDate.of(2022, 12, 1))) {
```
jonflynng, line 42:
```suggestion
    if (today.isBefore(LocalDate.of(2020, 12, 1))) {
```

jonflynng: Doesn't really matter but the tests are using 2020 so maybe makes it a bit more straightforward if someone new reads this.

jonflynng (approved):

jonflynng:
Looks good, there is one more thing to consider though. There are still some items in `SalesforcePriceRiseCreationComplete` which keep failing every day, so this will trigger the alarm as soon as this is merged. These fail due to insufficient fields retrieved from Salesforce. We can remove them from the DynamoDB table if needed, I can walk you through it.

PR #748, opened 2022-11-23
Fix engine picking up old Zone ABC rate plans

jonflynng:
Any Guardian Weekly subscriptions with a Zone A/B/C rate plan are migrated to the newer ROW and Domestic rate plans. We  first check if the subscription is currently on a Zone ABC rateplan. There was a mistake [here](https://github.com/guardian/price-migration-engine/pull/748/files#diff-c141291fe554e88a48be933c43e9c19f945d104eea3fb77d07d7af4ee8a09f0eL83) which was looking through the whole subscription history to find a Zone ABC rate plan, rather than just checking the active one. This means any subscriptions who used to be on a Zone ABC rate plan were being incorrectly processed. However, after checking the cohort, there were only a few subscriptions with this edge case, so little harm was done.

One line fix:
`val isZoneABC = subscription.ratePlans.filter(zoneABCPlanNames contains _.productName)`
to
`val isZoneABC = ratePlans.filter(zoneABCPlanNames contains _.productName)`


lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -162,7 +162,6 @@ object AmendmentData {
         .toRight(AmendmentDataFailure(s"Failed to get RatePlan for charges: $ratePlanCharges"))
 
       isZoneABC = zoneABCPlanNames contains ratePlan.productName
-      // check the Zone ABC rateplan is the currently active one -> effectiveEndDate property on the subscription, are we filtering out inactive Zone ABC rateplans in the bigquery query?? does the engine filter out inactive rateplans??
```
jonflynng, line 165: this `ratePlan` variable is the currently active one. There's a function above which matches rate plans on the subscription and the invoice which extracts the active plan.

shtukas (approved):
Great

lambda/src/test/scala/pricemigrationengine/model/ZuoraSubscriptionUpdateTest.scala
```diff
@@ -52,6 +52,98 @@ class ZuoraSubscriptionUpdateTest extends munit.FunSuite {
...
+            )
+          ),
+          currentTerm = None,
+          currentTermPeriodType = None
+        )
+      )
+    )
+  }
+
+  test("updateOfRatePlansToCurrent: is correct for a GW ROW subscription with a past Zone ABC rate plan") {
```
jonflynng, line 101: This is the test we needed

PR #756, opened 2022-12-13
Remove requirement for address.postalCode in the NotificationHandler

shtukas:
### What
We remove the requirement for a post code in the address of a customer about to be notified.

### Why
Turns out that not all addresses (notably outside the UK) have a post code. That requirement was preventing some subscriptions from being correctly process by the Notification Handler. 

( Interestingly we have had this problem for a long time, but those subscriptions were silently sitting in `SalesforcePriceRiseCreationComplete` stage, and now there comes a point where they become close to their `startDate` and hit the check introduced by this change https://github.com/guardian/price-migration-engine/pull/747 )


paulbrown1982 (approved):
👍  Do we also need to tweak any of the start dates?

shtukas:
@paulbrown1982 
Yes, I didn't mention that in the description but a few of those that were failing to to the startDate need to be pushed. 

@jonflynng 
Let's do that.

PR #761, opened 2023-02-01
account for rounding errors at the 20% cap in the amendment step

shtukas:
DataTech reported an error to us which came down to the following error in the price migration engine

```
AmendmentDataFailure(
    The new price 647.88 for cohort item: Some(CohortItem((...)(...),Some(GBP),Some(539.88),Some(575.88),(...)) (after amendment) was higher than the estimatedNewPrice (647.856)
)
```

It's unnecessary to fail for rounding errors, so this PR allows for them.


shtukas:
Merci Kelvin. I made a card to introduce a proper amendment handler test suite (we don't have one at the moment and I don't want to rush doing it now ^^)

PR #781, opened 2023-03-10
Reimplement price capping

shtukas:
This PR description is in two parts. The short version that just describes the change and the long version that Pascal had originally written with all the details (of the problem and solutions)

# Short Version

In this PR, we reimplement the 20% price capping policy which was introduced here: https://github.com/guardian/price-migration-engine/pull/727 , while also changing the semantics of the `estimatedNewPrice` field of a cohort Item (essentially reverting back to its original meaning). 

In one sentence: we no longer store the capped price in the `estimatedNewPrice` field. We store it uncapped and cap it when we need. The advantage is a much simpler capping logic and a more robust Amendment process (which requires the uncapped version). 

The Amendment handler now has a single `PriceCapper` object. It is not used in the Estimation handler. `PriceCapping.cappedPrice` is used just once in the SalesforcePriceRiseCreation handler and used also just once in the NotificationHandler. Then, at the very end, `PriceCapping.priceCorrectionFactor` is used once in the Amendment Handler (to compute the correct subscription update), and then `PriceCapping.cappedPrice` used one more time to check the price post subscription update (at which point a failure may be issued).

ps: The change looks big because it's actually 3 PRs into one. 

1. Decommissioning the old price capping.
2. Reimplementing price capping in a much simpler way.
3. Adapting the SalesforcePriceRiseCreation and the Notification handlers to the updated semantics of the `estimatedNewPrice` field.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
@@ -280,3 +280,23 @@ object AmendmentData {
     "Annual" -> 12
   )
 }
+
+object PriceCapper {
+
+  /*
+    This object implements the policy of not increasing subscription prices, and
+    therefore what our customers pay, by more then 20% during a single price rise.
+   */
+
+  private val priceCappingMultiplier = 1.2 // old price + 25%
+  def cappedPrice(oldPrice: BigDecimal, estimatedNewPrice: BigDecimal): BigDecimal =
+    List(estimatedNewPrice, oldPrice * priceCappingMultiplier).min
+
+  def priceCorrectionFactor(oldPrice: BigDecimal, estimatedNewPrice: BigDecimal): BigDecimal = {
+    if (estimatedNewPrice == 0 || estimatedNewPrice < oldPrice * priceCappingMultiplier) {
+      1
+    } else {
+      (oldPrice * priceCappingMultiplier) / estimatedNewPrice
+    }
+  }
+}
--- a/lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
@@ -1,6 +1,5 @@
 package pricemigrationengine.model
 
-import pricemigrationengine.model.ChargeCap.ChargeCapBuilderFromMultiplier
 import pricemigrationengine.model.ZuoraProductCatalogue.{homeDeliveryRatePlans, productPricingMap}
 
 import java.time.LocalDate
@@ -35,11 +34,10 @@ object AmendmentData {
       subscription: ZuoraSubscription,
       invoiceList: ZuoraInvoiceList,
       earliestStartDate: LocalDate,
-      chargeCapBuilderOpt: Option[ChargeCapBuilderFromMultiplier]
   ): Either[AmendmentDataFailure, AmendmentData] = {
     for {
       startDate <- nextServiceStartDate(invoiceList, subscription, onOrAfter = earliestStartDate)
-      price <- priceData(account, catalogue, subscription, invoiceList, startDate, chargeCapBuilderOpt)
+      price <- priceData(account, catalogue, subscription, invoiceList, startDate)
     } yield AmendmentData(startDate, priceData = price)
   }
 
@@ -76,8 +74,7 @@ object AmendmentData {
       catalogue: ZuoraProductCatalogue,
       subscription: ZuoraSubscription,
       invoiceList: ZuoraInvoiceList,
-      startDate: LocalDate,
-      chargeCapBuilderOpt: Option[ChargeCapBuilderFromMultiplier]
+      startDate: LocalDate
   ): Either[AmendmentDataFailure, PriceData] = {
 
     def hasNotPriceAndDiscount(ratePlanCharge: ZuoraRatePlanCharge) =
@@ -147,17 +144,13 @@ object AmendmentData {
       currency <- pairs.headOption
         .map(p => Right(p.chargeFromSubscription.currency))
         .getOrElse(Left(AmendmentDataFailure(s"No invoice items for date: $startDate")))
-      oldPrice <- totalChargeAmount(subscription, invoiceList, startDate, None)
-      newPriceWithoutCapping <- totalChargeAmount(pairs)
-      newPriceWithCapping = chargeCapBuilderOpt match {
-        case None                   => newPriceWithoutCapping
-        case Some(chargeCapBuilder) => List(newPriceWithoutCapping, chargeCapBuilder(oldPrice).priceCap).min
-      }
+      oldPrice <- totalChargeAmount(subscription, invoiceList, startDate)
+      newPrice <- totalChargeAmount(pairs)
       billingPeriod <- pairs
         .flatMap(_.chargeFromSubscription.billingPeriod)
         .headOption
         .toRight(AmendmentDataFailure("Unknown billing period"))
-    } yield PriceData(currency, oldPrice, newPriceWithCapping, billingPeriod)
+    } yield PriceData(currency, oldPrice, newPrice, billingPeriod)
   }
 
   /** Total charge amount, including taxes and discounts, for the service period starting on the given service start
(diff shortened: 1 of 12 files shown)


shtukas:
# Long version

## 1. What is the problem ?

As it is currently implemented, the 20% price capping is incorrect. The effect of that problem over the past 5 months has been that a few items have failed the price capping check at the end of the Amendment step. Every time it was possible to correct them manually, but a recent failure of an annual Voucher has led to a proper investigation whose conclusion is that the price capping is wrongly implemented. 

The problem manifests itself when there are, multiple charges in one payment operation. Here is a simple example to illustrate it:

A monthly subscription works like this

```
Jan: User pays £10
Feb: User pays £10
...
Dec: User pays £10
```

If we price rise them, and the new price would be £13 then the engine will apply its cap and the user will pay £12 monthly. 

For annuals things are different. An annual can look like this

```
Sub product 1: £100
Sub product 2: £100
```

The user pays 200£ for the year. Now when we price rise this, the new price could be 

```
Sub product 1: £130
Sub product 2: £130
```

Which corresponds to a 30% increase and a total of £260. But the engine estimation estimation handler will compare each sub product price of £130 with the previous total +20%, meaning compares £130 with £240, and will think that everything is absolutely fine. But then the engine will move to amendment, and then will compare the new total of £260 with £240 and will error extremely confused that the 20% increase cap had not been applied. 

## 2. What is the solution ?

The solution is simply to compare total prices and total prices, not total prices and charges. Total prices and charges tend to be equal for monthlies but not for quarterlies and annuals. 

If we see that the new price is greater than the old price +20% we know that the new price needs to be capped. 

The capping here is simpler than in the previous implementation. We compute a correcting factor that moves the new (uncapped) price to its capped value and apply that factor to the individual charges.  

For instance, in the previous example of an annual paying £200, 

```
Sub product 1: £100
Sub product 2: £100
```

which would be increased to £260

```
Sub product 1: £130
Sub product 2: £130
```

Then we compute the correcting factor as `240/260 = 0.923076` and apply that to each component of the price, leading to 

```
Sub product 1: £119.99
Sub product 2: £119.99
```

## 3. So what's the deal with re-interpreting the `estimatedNewPrice` ?

At the moment, the estimated new price is capped during the Estimation step. And that price is used during the various notifications (Salesforce and Braze) happening between the Estimation step and the Amendment step. That's nice and convenient, but it causes a problem. 

When it's time for the Amendment step, we need the correcting factor **before** amending the subscription, but then how do we know the correcting factor without knowing what the total (uncapped price) price is going to be (or would have been) ? (A classical _chicken and egg_ problem.) This is where the estimated new price should never have been capped in the database. 

So in this change we do this. 

- When we run the Estimation step, we compute the natural uncapped price, and we store it uncapped. 

- When we update Salesforce, we cap the price `min(oldPrice * 1.2, estimatedNewPrice)`

- When we send the notification to the user (email or letter), we cap the price in that communication (using the same formula)

- When we are about to amend the subscription in Zuora, we know (using the old price and the uncapped estimated price) what is the correct correcting factor to applies to the charges. This value is either 1 if the price didn't need to be capped, and otherwise is less than one (for instance 0.923076 above). When we compute the subscription update object with that correcting factor, then the price of the subscription is Zuora automatically result in the capped price.

- Obviously we can also check after that the subscription has been updated that the price is the capped price (a check we are currently doing).  

The price capping is now much easier to follow and understand. The Amendment Handler now has a single `PriceCapper` object. It is not used in the Estimation handler. `PriceCapping.cappedPrice` is used just once in the SalesforcePriceRiseCreation handler and used also just once in the NotificationHandler. Then, at the very end, `PriceCapping.priceCorrectionFactor` is used once in the Amendment Handler (to compute the correct subscription update), and then `PriceCapping.cappedPrice` used one more time to check the price post subscription update (at which point a failure may be issued).

### In which exact circumstances can the post subscription update fail ?

Answer:

- If the subscription changes between the Estimation step and the Amendment step *and* the `estimatedNewPrice` increases above the capped Price (or was already above it but raises even higher), then the post subscription update price check will fail. (The course of action in this case, is simply to clean the subscription by manually removing the price rise engine amendments in Zuora and put the subscription back to `ReadyForEstimation` processing stage , at which point the correct/updated `estimatedNewPrice` will be computed.)

- If the Zuora subscription update object is wrong and brings the price above the intended capped price, then the check will fail. (The course of action here is to investigate and correct the bug in the code.)

## 4. Is this backward compatible ?

If we commit this change, the engine will consider that the estimated prices of existing items in the database are the uncapped prices. This could be untrue, and in some circumstances the engine _could_ fail if the final price is too high, due to the post Amendment check (and we know we can easily recover from that). (Intuitively, it's like the Engine would have been led, by this PR, to _believe_ that a few rare items in the database didn't need a cap, to discover after amendment that the true estimated uncapped price wasn't what was recorded). 

But then, even the above would be a better situation than we are in. In the current situation we have a few failures per quarter (that's admittedly not a lot) but this will last for ever because the problem is intrinsic to the engine itself. In the new situation we might have a few failures in the near future (and again we can easily recover from it), but the discrepancies that causes them will eventually disappear from the engine as the data corrects itself, and in particular **future price migrations will not suffer from it**. 

## 5. Alternative solutions

There are two alternative solutions, to the above. I am mentioning them for completion

1. We could expand the Dynamo table schema to have one field for the capped estimated price and one for the uncapped one. The capped one to be used in communications and the uncapped one used during the Amendment step to compute the price correction factor.

2. We could also as first step of the Amendment step run a process similar to the one that had been run during the Estimation step just to compute the uncapped price.

Neither of those solutions was selected because they are more complex than the more natural solution of not capping the stored estimated price in the first place.

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -315,3 +280,23 @@ object AmendmentData {
...
+    This object implements the policy of not increasing subscription prices, and
+    therefore what our customers pay, by more then 20% during a single price rise.
+   */
+
+  private val priceCappingMultiplier = 1.2 // old price + 25%
+  def cappedPrice(oldPrice: BigDecimal, estimatedNewPrice: BigDecimal): BigDecimal =
+    List(estimatedNewPrice, oldPrice * priceCappingMultiplier).min
+
+  def priceCorrectionFactor(oldPrice: BigDecimal, estimatedNewPrice: BigDecimal): BigDecimal = {
+    if (estimatedNewPrice == 0 || estimatedNewPrice < oldPrice * priceCappingMultiplier) {
```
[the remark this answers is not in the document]
shtukas: Updated. Thanks!

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -293,7 +293,7 @@ object PriceCapper {
     List(estimatedNewPrice, oldPrice * priceCappingMultiplier).min
 
   def priceCorrectionFactor(oldPrice: BigDecimal, estimatedNewPrice: BigDecimal): BigDecimal = {
-    if (estimatedNewPrice == 0 || estimatedNewPrice < oldPrice * priceCappingMultiplier) {
+    if (estimatedNewPrice == 0 || estimatedNewPrice.compareTo(oldPrice * priceCappingMultiplier) <= 0) {
```
kelvin-chappell, line 296:
```suggestion
    if (estimatedNewPrice == 0 || estimatedNewPrice.compareTo(oldPrice.multiply(BigDecimal(priceCappingMultiplier))) <= 0) {
```

shtukas: Screenshot 2023-03-16 at 11 30 43

shtukas: @kelvin-chappell

kelvin-chappell: I was looking at `java.math.BigDecimal` rather than `scala.math.BigDecimal`.
The former has a [multiply](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/math/BigDecimal.html#multiply%28java.math.BigDecimal%29) method but the latter uses [*](https://www.scala-lang.org/api/2.12.6/scala/math/BigDecimal.html#*(that:scala.math.BigDecimal):scala.math.BigDecimal) for the same thing.
So I withdraw this suggestion, as it's redundant.

kelvin-chappell (approved):
Great work! 👍

shtukas:
Merci @kelvin-chappell

PR #784, opened 2023-03-13
test

shtukas:


lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -99,16 +104,20 @@ object AmendmentHandler extends CohortHandler {
...
-      item.subscriptionName,
-      startDate,
-      oldPrice,
-      newPrice,
-      estimatedNewPrice,
-      newSubscriptionId,
-      whenDone
-    )
+    } yield {
+      println(s"update: ${update}")
```
kelvin-chappell, line 110: You can always do `.debug("update")` on line 84 instead for an ad hoc inspection
https://zio.dev/guides/tutorials/debug-a-zio-application/#debugging-a-zio-application-using-debug-effect

shtukas: Thanks @kelvin-chappell !

PR #791, opened 2023-03-17
correct subscription ids file upload

shtukas:
The SubscriptionIdUploadHandler is skipping the first subscriptionId that is incorrectly read as a csv header. The salesforce-subscription-id-report.csv file (as well as the excluded-subscription-ids.csv file) only contains subscriptionIds so that was a mistake. This change corrects it.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/SubscriptionIdUploadHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/SubscriptionIdUploadHandler.scala
@@ -12,7 +12,7 @@ import scala.jdk.CollectionConverters._
 
 object SubscriptionIdUploadHandler extends CohortHandler {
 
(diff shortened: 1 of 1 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/SubscriptionIdUploadHandler.scala
```diff
@@ -12,7 +12,7 @@ import scala.jdk.CollectionConverters._
 
 object SubscriptionIdUploadHandler extends CohortHandler {
 
-  private val csvFormat = CSVFormat.Builder.create().setHeader().setSkipHeaderRecord(true).build()
+  private val csvFormat = CSVFormat.Builder.create().setHeader().build()
```
kelvin-chappell, line 15: I wonder if you need to get rid of `setHeader` as well.  Perhaps a test would help.

shtukas:
See Part 2: https://github.com/guardian/price-migration-engine/pull/792

PR #792, opened 2023-03-17
correct subscription ids file upload (part 2)

shtukas:
This is Part 2 of https://github.com/guardian/price-migration-engine/pull/791

The change in Part 1 had no effect whatsoever, but I didn't notice it because I didn't know there was a test. 

This change actually implements the change, and the test was updated accordingly 🎉


kelvin-chappell (approved):
There's always a test! 👍

shtukas:
💜

PR #798, opened 2023-03-18
Implement membership price rise

shtukas:
Building on the recent reimplementation of price rise ( https://github.com/guardian/price-migration-engine/pull/781 ) , together with allowing for overriding it ( https://github.com/guardian/price-migration-engine/pull/797 ), we introduce the membership price migration.

In particular:

- The `CohortSpec` has been updated with function `isMembershipPriceRiseBatch1` to know if we are running that migration. (Note that the membership price migration will go in three batches.)

- The signature of the `main` functions of the EstimationHandler, SalesforcePriceRiseCreationHandler, NotificationHandler and AmendmentHandler has been updated to consume the `CohortSpec`. That information is passed down the chain of command and is used to determine whether to override price capping as well as which pricing method to perform.

- `AmendmentData.priceData`  now supports the old style print migrations and the new (simple) membership migration. What prompted this extension is the recent discovery of subscriptions without an active matching product rate plan (which was causing an error during the Estimation step). 

- The rest of the change is just updating the tests. Speaking of tests, changes in the tests are of two kinds:  (1) adapting to the signature update of the handlers `main` function, and (2) ensuring that we test both the non membership versus membership versions of various operations.

Tests (and fixtures) are written for GBP and AUD.


shtukas:
Note: 

If it had been only to avoid price capping membership subscriptions, this PR would have been much simpler and would have used a hack in one form or another, but the discovery that the existing print products dependant pricing logic was breaking for membership subscriptions could only be handled by parameterising the step handlers with cohort spec data. Of course, with that done, the price capping can then be driven more subtly (turned on or off depending on the cohort as we do here).

Interestingly this then opens the possibility, if there is such a need for it in the future, to drive more subtle behaviours of the engine per migration. One possibility would simply be to expand the CohortSpec object itself. This is a part of the data model that is easy to expand, and yet can lead to dramatically powerful variations of the engine and migrations in general.

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -63,18 +67,54 @@ object AmendmentData {
...
+      nextServiceStartDate: LocalDate,
+      cohortSpec: CohortSpec,
+  ): Either[AmendmentDataFailure, PriceData] = {
+
+    // Note: The nextServiceDate has gone through:
+    // cohortSpec.earliestPriceMigrationStartDate >> `spreadEarliestStartDate` >> `nextServiceStartDate`
+
+    /*
+      Date: March 2023
+      Author: Pascal
```
kelvin-chappell, line 84: Isn't this going to be clear from the history?

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -63,18 +67,54 @@ object AmendmentData {
...
+
+      With the introduction of the Membership price migration there is now two ways to compute PriceData: The old way,
+      implemented in priceDataWithRatePlanMatching, which was used for the ore complex print subscriptions, and
```
kelvin-chappell, line 87:
```suggestion
      implemented in priceDataWithRatePlanMatching, which was used for the more complex print subscriptions, and
```

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
```diff
@@ -63,18 +67,54 @@ object AmendmentData {
...
+      a simplified way that we are going to apply to membership.
+
+      This split came from the fact that at the time these lines are written, it is not possible to find matching
+      product rate plan charges for for the subscription rate plan charges because the previous, pre migration,
```
kelvin-chappell, line 91:
```suggestion
      product rate plan charges for the subscription rate plan charges because the previous, pre migration,
```

lambda/src/main/scala/pricemigrationengine/model/AmendmentData.scala
shtukas: In many projects I tend to date and sign important comments, above all if they might come across as a surprise to people who look at the code and without having seen the PR. Maybe not as important in this project actually. In any case, I will probably remove the entire section when I come back doing more refactoring :)

shtukas:
Amazing. Thanks @kelvin-chappell . The engine is a bit overwelminjg at first, but as far as I am concerned, it's by far the best well thought of and well written project of its size at the Guardian. The tests, the isolation, the purity, the effect system, the modularity, all contribute to make daring adaptations and refactorings completely achievable (as long as one is familiar enough with the logic).

@paulbrown1982

PR #801, opened 2023-03-20
Prevent notification step in case of membership migration

shtukas:
This will allow the run of the state machine up to the Salesforce updates, but without the Braze notifications, in the case of the membership migration. (The Notification handler would fire up during state machine run, but we don't want that, yet)


kelvin-chappell (approved):
👍

PR #822, opened 2023-04-01
Add support for membership Batch2

shtukas:
We add support for membership Batch2. The two batches are identical in structure with the exception of a month delay. Consequently the code change is trivial, but for symmetry the tests have been mirrored for Batch2, which makes most of the change.


kelvin-chappell (approved):
The code comments are very good.  Perhaps also a readme for estimation and amendment might help to summarise the difference between the two processes and how the timing of it works, for someone new to the codebase who's trying to work their way around it.

PR #823, opened 2023-04-01
Better Notification timing tests

shtukas:
Out of perfectionism we edit the notification timing tests for membership. More exactly:

- Use the real life notification dates.
- Better explain and better test the notification window.


kelvin-chappell (approved):
I wonder if it might also be worth including a readme file to give an overview of the notification rules?  It can be difficult to follow the overall logic of it from the code comments.

shtukas:
Good idea!

PR #827, opened 2023-04-04
set scala stewart frequency to weekly

shtukas:
The frequent AWS updates can be a bit noisy. Setting instructions for Scala Stewart to run weekly. ( documentation: https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md )


kelvin-chappell:
You can also group updates into a single PR like this if it helps: https://github.com/guardian/identity/pull/2380

PR #834, opened 2023-04-06
Better implementation of Estimation lead time check

shtukas:
This PR is a refinement and better implementation of the check we installed here: https://github.com/guardian/price-migration-engine/pull/829

The original implementation mostly didn't consider that at cohort creation time the items do not yet have a `startDate`, which would lead the check to fail. 

Here we replace the check with a computation (which is a huge win). We ensure that at Estimation stage the startDate is legal (from a user notification requirement point of view), meaning sufficiently ahead in the future relatively to today, and also after the cohort's earliestPriceMigrationStartDate.

To do this we introduce `decideEarliestStartDate` which adds rationality to `spreadEarliestStartDate`, together with the help of `datesMax`. Those two functions are then what is being tested. 

(In the previous PR we had pointed out that `thereIsEnoughNotificationLeadTime` was then being tested twice. It is back being tested once as part of the Notification handler's tests. In this situation we have better computation during Estimation and the same original check during Notification, which logically makes more sense).

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
@@ -1,8 +1,8 @@
 package pricemigrationengine.handlers
 
-import pricemigrationengine.handlers.NotificationHandler.thereIsEnoughNotificationLeadTime
+import pricemigrationengine.handlers.NotificationHandler.minLeadTime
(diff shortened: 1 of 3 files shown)


shtukas:
Screenshot 2023-04-07 at 19 01 14

shtukas:
Screenshot 2023-04-07 at 19 02 59

shtukas:
Screenshot 2023-04-07 at 19 06 02

PR #838, opened 2023-04-07
checkExpirationTiming in Amendment handler

shtukas:
We have a subscription with an effective end date of `2023-04-10`, corresponding to a cohort item with a startDate of `2023-05-10`. 

04 Screenshot 2023-04-07 at 19 57 08

Screenshot 2023-04-07 at 21 34 11

This caused the following error in Zuora: 

```
The Contract effective date should not be later than the term end date of the basic subscription
```

This is the result of the relevant check not having been performed during the Estimation step.

This change is the first of two changes. Here we add a check to the Amendment handler to move that item to `Cancelled` state. The next change will add that check to the Estimation handler. (We should indeed do this during Estimation to avoid unnecessarily notify the corresponding customers).

 The offending item has been put to `Cancelled` state.
  
11 Screenshot 2023-04-07 at 21 18 36

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -73,7 +73,7 @@ object AmendmentHandler extends CohortHandler {
         )
       )
 
-  private def checkExpirationTiming(
+  def checkExpirationTiming(
       item: CohortItem,
       subscription: ZuoraSubscription
   ): Either[Failure, Unit] = {
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -39,12 +39,24 @@ object AmendmentHandler extends CohortHandler {
           val result = CancelledAmendmentResult(item.subscriptionName)
           CohortTable.update(CohortItem.fromCancelledAmendmentResult(result)).as(result)
         }
+        case _: ExpiringSubscriptionFailure => {
+          // `ExpiringSubscriptionFailure` happens when the subscription's end of effective period is before the
+          // intended startDate (the price increase date). Alike `CancelledSubscriptionFailure` we cancel the amendment
+          // and the only effect is an updated cohort item in the database
+          val result = ExpiringSubscriptionResult(item.subscriptionName)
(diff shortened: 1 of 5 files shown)


shtukas:
It's interesting that this was not caught during Estimation step, but the problem, as I have just realised, is that Zuora is happy generating invoice previews for subscriptions passed their expiration date. Giving the false impression that the subscription will be alive..., in any case `nextServiceStartDate` doesn't fail there...

Here is an example, this subscription expires in a few days (`2023-04-10`)

```
{
  "success" : true,
  "id" : "[REMOVED]",
  "accountId" : "[REMOVED]",
  "accountNumber" : "[REMOVED]",
  "accountName" : "[REMOVED]",
  "invoiceOwnerAccountId" : "[REMOVED]",
  "invoiceOwnerAccountNumber" : "[REMOVED]",
  "invoiceOwnerAccountName" : "[REMOVED]",
  "subscriptionNumber" : "A-S00663936",
  "version" : 5,
  "revision" : "5.0",
  "termType" : "TERMED",
  "invoiceSeparately" : false,
  "contractEffectiveDate" : "2018-04-10",
  "serviceActivationDate" : "2018-04-10",
  "customerAcceptanceDate" : "2018-04-10",
  "subscriptionStartDate" : "2018-04-10",
  "subscriptionEndDate" : "2023-04-10",
  "lastBookingDate" : "2022-04-10",
  "termStartDate" : "2022-04-10",
  "termEndDate" : "2023-04-10",
  "initialTerm" : 12,
   (...)
```

but the Zuora API was happy invoicing it for as long as I wanted, notably

```
{
    "id" : "064acf5fb17349bdb41f0ddb7d9fcc79",
    "subscriptionName" : "A-S00663936",
    "subscriptionId" : "8a12870d8755735e01875dbf871a48d1",
    "subscriptionNumber" : "A-S00663936",
    "serviceStartDate" : "2023-12-10",
    "serviceEndDate" : "2024-01-09",
    "chargeAmount" : 5.000000000,
    "chargeDescription" : "",
    "chargeName" : "Supporter Membership - Monthly",
    "chargeNumber" : "C-01285336",
    "chargeId" : "2c92a0ff62aed7e60162af285d332aeb",
    "productName" : "Supporter",
    "quantity" : 1,
    "taxAmount" : 0.000000000,
    "unitOfMeasure" : "",
    "chargeDate" : "2023-04-07 23:04:06",
    "chargeType" : "Recurring",
    "processingType" : "Charge",
    "appliedToItemId" : null
  }
```

Although here, it seems to compute taxes differently...

PR #852, opened 2023-04-26
Introduce EngineSecrets

shtukas:
### Context

At the moment we are using the following pattern to set environment variables

```
salesforceAuthUrl:
  !Sub
  - '{{resolve:secretsmanager:price-migration-engine-lambda-${Stage}:SecretString:salesforceAuthUrl::${SecretsVersion}}}'
  - SecretsVersion: !FindInMap [StageMap, !Ref Stage, SecretsVersion]
```

from which we retrieve secrets. 

### Problem

We (DevEx) recently discovered that it causes credentials leaks in AWS logs. 

### Solution

We are migrating to calling SecretsManager programatically.  We update the Cloud Formation to allow for calling SecretsManager and introduce the service `SecretManagerClient`, and decommission the leak from the cloud formation

### Observation

Because we are now extracting the entire json objects in the code, we no longer have the very tight retrieval we had. The same handlers as before retrieve secrets, but they now essentially have access to the entire object.


shtukas:
Already tested in CODE

gnm-notona (approved):
Looks great, only general question about why we aren't using parameter store?

shtukas:
Great question. 
1. This decision predates me, so I am just running with it.
2. I noticed that people in Supporter Revenue seem to be more familiar with Secrets Manager than Parameter Store.

project/Dependencies.scala
```diff
@@ -8,6 +8,8 @@ object Dependencies {
   lazy val awsS3 = "software.amazon.awssdk" % "s3" % awsSdkVersion
   lazy val awsSQS = "software.amazon.awssdk" % "sqs" % awsSdkVersion
   lazy val awsStateMachine = "software.amazon.awssdk" % "sfn" % awsSdkVersion
+  lazy val awsSecretsManager = "software.amazon.awssdk" % "secretsmanager" % awsSdkVersion
+  lazy val playJson = "com.typesafe.play" %% "play-json" % "2.9.4"
```
kelvin-chappell, line 12: There's already a Json-handling library in the repo, called `upickle`, so this looks to be redundant.

lambda/src/main/scala/pricemigrationengine/services/PriceMigrationEngineSecrets.scala
```diff
@@ -0,0 +1,47 @@
...
+  implicit val secretsRead = Json.reads[PriceMigrationEngineSecrets]
+}
+
+object SecretManagerClient {
+
+  lazy val region: regions.Region = regions.Region.EU_WEST_1
+
+  lazy val secretsClient = SecretsManagerClient
+    .builder()
+    .build()
```
kelvin-chappell, line 39:
```suggestion
  lazy val secretsClient = SecretsManagerClient.create()
```

lambda/src/main/scala/pricemigrationengine/services/PriceMigrationEngineSecrets.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  lazy val secretId: String = "price-migration-engine-lambda-CODE"
```
kelvin-chappell, line 41: Should pass `stage` name in as an env value

lambda/src/main/scala/pricemigrationengine/services/PriceMigrationEngineSecrets.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  lazy val secretsJsonString =
+    secretsClient.getSecretValue(GetSecretValueRequest.builder().secretId(secretId).build()).secretString()
```
kelvin-chappell, line 44: Anything that relies on an external service should really be in a ZIO service or wrapped in a ZIO effect.  Unless this is just experimental

kelvin-chappell (commented):
I think I would model it slightly differently and have a `Configuration` service with a `SecretManager` implementation.  But that's just a matter of opinion and there are many different ways of modelling it.
Other than making sure all effects are wrapped in ZIOs it looks fine.

lambda/src/main/scala/pricemigrationengine/services/EnvConfig.scala
```diff
@@ -21,9 +23,9 @@ object EnvConfig {
   object zuora {
     val layer: Layer[ConfigFailure, ZuoraConfig] = ZLayer.fromZIO(
       for {
-        zuoraApiHost <- env("zuoraApiHost")
-        zuoraClientId <- env("zuoraClientId")
-        zuoraClientSecret <- env("zuoraClientSecret")
+        zuoraApiHost <- ZIO.succeed(SecretManagerClient.secrets.zuoraApiHost)
```
kelvin-chappell, line 26: This shouldn't be a `ZIO.succeed` unless there's absolutely no possibility of it failing

lambda/src/main/scala/pricemigrationengine/services/PriceMigrationEngineSecrets.scala
shtukas: Done!

lambda/src/main/scala/pricemigrationengine/services/EngineSecrets.scala
```diff
@@ -0,0 +1,56 @@
...
+
+object EngineSecrets {
+
+  implicit val reader: Reader[EngineSecrets] = macroRW
+
+  private lazy val region: regions.Region = regions.Region.EU_WEST_1
+
+  private lazy val secretsClient = SecretsManagerClient.create()
+
+  private lazy val secretId: String = s"price-migration-engine-lambda-${getenv("stage")}"
```
kelvin-chappell, line 32: `getenv` is really an effect and there's a ZIO wrapper for it: https://zio.dev/reference/services/system/#system-environment

lambda/src/main/scala/pricemigrationengine/services/EngineSecrets.scala
```diff
@@ -0,0 +1,56 @@
...
+  def getSecretString: ZIO[Any, ConfigFailure, String] = {
+    for {
+      s <- ZIO
+        .attempt(
+          secretsClient.getSecretValue(GetSecretValueRequest.builder().secretId(secretId).build()).secretString()
+        )
+        .mapError { ex =>
+          ConfigFailure(s"Failure to retrieve secrets string: ${ex.getMessage}")
+        }
+    } yield s
```
kelvin-chappell, line 43:
```suggestion
    ZIO
        .attempt(
          secretsClient.getSecretValue(GetSecretValueRequest.builder().secretId(secretId).build()).secretString()
        )
        .mapError { ex =>
          ConfigFailure(s"Failure to retrieve secrets string: ${ex.getMessage}")
        }
    }
```

lambda/src/main/scala/pricemigrationengine/services/EngineSecrets.scala
shtukas: Nice one. I was wondering about that...

shtukas: Done here: https://github.com/guardian/price-migration-engine/pull/852/commits/be8cb7e5f55bbbb1e177b00bd68db916379adb7d

PR #897, opened 2023-07-12
Introduce MigrationType

shtukas:
When we introduced the 2023 membership price migrations, we introduced the concept of parametrisation of basic functions with `CohortSpec`s. Back then we got away with `if (cohort detections) then  ... else (else if ...)` the existing code, but with more of such customised migrations coming, it was time for a refactoring. 

For this we introduce `sealed trait MigrationType`, leading to a much cleaner and more readable code. 

`MigrationType` does not identity a migration (Sometime several migrations map to a unique migration type, for instance both Membership 2023 Batch 1 and Batch 2 both map to type `Membership2023Monthlies`, whereas Membership 2023 Batch 3 identifies with `Membership2023Annuals`.  It simply helps identify common code used by possibly more than one migration.

All pre 2023 migrations map to type `Legacy`.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -19,28 +19,31 @@ object NotificationHandler extends CohortHandler {
   // For general information about the notification period see the docs/notification-periods.md
 
   // The standard notification period for letter products (where the notification is delivered by email)
-  // is -49 (included) to -35 (excluded) days.
-  val guLettersNotificationLeadTime = 49
-  private val engineLettersMinNotificationLeadTime = 35
+  // is -49 (included) to -35 (excluded) days. Legally the min is 30 days, but we set 35 days to alert if a
+  // subscription if exiting the notification window and needs to be investigated and repaired before the deadline
+  // of 30 days.
+
+  val letterMaxNotificationLeadTime = 49
+  private val letterMinNotificationLeadTime = 35
 
   // Membership migration
   // Notification period: -33 (included) to -31 (excluded) days
-  private val membershipPriceRiseNotificationLeadTime = 33
-  private val membershipMinNotificationLeadTime = 31
+  private val emailMaxNotificationLeadTime = 33
+  private val emailMinNotificationLeadTime = 31
 
   def maxLeadTime(cohortSpec: CohortSpec): Int = {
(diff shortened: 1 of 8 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -156,8 +159,13 @@ object NotificationHandler extends CohortHandler {
       currencyISOCode <- requiredField(cohortItem.currency, "CohortItem.currency")
       currencySymbol <- currencyISOtoSymbol(currencyISOCode)
 
-      // In the case of membership price rise, we need to not cap the price
-      cappedEstimatedNewPriceWithCurrencySymbol = s"${currencySymbol}${PriceCap.cappedPrice(oldPrice, estimatedNewPrice, CohortSpec.isMembershipPriceRise(cohortSpec))}"
+      forceEstimated = MigrationType(cohortSpec) match {
+        case Membership2023Monthlies => true
+        case Membership2023Annuals   => true
+        case _                       => false
+      }
```
kelvin-chappell, line 166: Not sure if this is an improvement or not:
```suggestion
      forceEstimated = MigrationType(cohortSpec) == Membership2023Monthlies ||  MigrationType(cohortSpec) == Membership2023Annuals
```

lambda/src/main/scala/pricemigrationengine/handlers/SalesforcePriceRiseCreationHandler.scala
```diff
@@ -80,16 +80,21 @@ object SalesforcePriceRiseCreationHandler extends CohortHandler {
...
-      ), // In case of membership price rise, we override the capping
-      Some(priceRiseDate),
-      Some(subscription.Id)
-    )
+    } yield {
+      val forceEstimated = MigrationType(cohortSpec) match {
+        case Membership2023Monthlies => true
+        case Membership2023Annuals   => true
+        case _                       => false
+      }
```
kelvin-chappell, line 88: This is same as above.  Could possibly be extracted

kelvin-chappell (approved):
This is a great idea and makes much more readable code.  👍 
I had a trivial suggestion for one piece of code, which you can ignore if you think it doesn't help.

shtukas:
@kelvin-chappell 
Amazing! Thanks!

> I had a trivial suggestion for one piece of code

I am going to leave it as it is, because I am just about to add a new migration type, and will refactor (make those simplifications) once the new one is in :)

PR #896, opened 2023-07-12
Supporter+ 2023 Migration V1 to V2 

shtukas:
Implementation of the supporter plus migration from version 1 to version 2, essentially a change of rate plans, without price increase. 

This migration is different from the previous ones because there indeed in no price increase (with that said there is still emails to be sent to inform the subscribers of a change in terms and conditions) and also because the target rate plan has two charges instead of one.

Considering the slightly more complex logic, we have here thoroughly tested 4 subscriptions

1. Monthly Standard: a monthly subscription with a charge of 10 GBP, meaning doesn't have an extra contribution charge. The two important things to verify are how it evaluates and the amendment data. 

We evaluate as

```
PriceData(currency, 10, 10, "Month")

SuccessfulEstimationResult(
  subscriptionName = "SUBSCRIPTION-NUMBER",
  startDate = LocalDate.of(2023, 9, 1),
  currency = "GBP",
  oldPrice = BigDecimal(10.0),
  estimatedNewPrice = BigDecimal(10),
  billingPeriod = "Month"
)
```

which shows that prices before and after equal the rate plan(s). Note that in any other migration, the subscription would be given the processing stage `NoPriceIncrease`, but we modified that rule just for this migration. 

The update is then

```
ZuoraSubscriptionUpdate(
   add = List(AddZuoraRatePlan("8a128ed885fc6ded01860228f77e3d5a", LocalDate.of(2023, 8, 1))),
   remove = List(RemoveZuoraRatePlan("8a129d388962fe000189679374ce42a1", LocalDate.of(2023, 8, 1))),
   currentTerm = None,
   currentTermPeriodType = None
)
```

2. Monthly with Contribution. Here we have a 10 GBP rate plan with a 15 USD contribution totalling to 25 USD. 

We evaluate at 

```
PriceData(currency, 25, 10, "Month")

SuccessfulEstimationResult(
  subscriptionName = "SUBSCRIPTION-NUMBER",
  startDate = LocalDate.of(2023, 9, 3),
  currency = "GBP",
  oldPrice = BigDecimal(25.0),
  estimatedNewPrice = BigDecimal(10),
  billingPeriod = "Month"
)
```

The "old" price is correctly computed as 25 GBP (which is all that the customer is paying) , but note here that the new price is 10 GBP and is the charge of the new rate plan itself without contribution. 

The update is then

```
ZuoraSubscriptionUpdate(
  add = List(
    AddZuoraRatePlan(
      "8a128ed885fc6ded018602296ace3eb8",
[...]

PR #906, opened 2023-07-23
supporter plus: user data fix

shtukas:
The `sendNotification` function which was made during the era of notifications for print products has user data requirements that are not met by many digital products, and in this case doesn't find postal mailing data for the supporter plus migration ( https://github.com/guardian/price-migration-engine/pull/896 ). 

```
Failed to send price rise notification: 
    - NotificationHandlerFailure(Contact.MailingAddress is a required field)
    - NotificationHandlerFailure(Contact.OtherAddress.street is a required field)
    - NotificationHandlerFailure(Contact.OtherAddress.country is a required field
```

Here we relax that requirement to allow for the emails to be sent. Note that as we did for membership, we default to country UK so that the Braze canvas works.


shtukas:
When membership was set up we had the same process of updating the `sendNotification` function to relax user data requirements. Interestingly the engine fails silently when a notification could not be sent due to user data and we only discover the problem when the notification is exiting its notification window. We could make the engine error on such missing user data cases, to trigger an investigation earlier.

PR #912, opened 2023-07-31
Renewal before amendment when needed

shtukas:
When performing the amendment step, it can happen that the startDate of the new rate plan (which is meant to coincide with the next billing operation), falls after the subscription end of term date, in which case Zuora fails the amendment with error message:

```
The Contract effective date should not be later than the term end date of the basic subscription
``` 

The solution is simply to renew the subscription before performing the amendment, which can be done using the Zuora API. 

```
curl -i -X PUT \
  'https://rest.zuora.com/v1/subscriptions/A-S00885464/renew' \
  -H 'Authorization: Bearer [removed]' \
  -H 'Content-Type: application/json' \
  -d '{}'
```

Note: I have set the batch size to 100 (from 150), after noticing that at 150 the updated lambda ran for 12 minutes, too close for comfort to the AWS 15 mins limit.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -12,7 +12,7 @@ import java.time.LocalDate
 object AmendmentHandler extends CohortHandler {
 
   // TODO: move to config
-  private val batchSize = 150
+  private val batchSize = 100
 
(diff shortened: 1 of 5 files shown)


shtukas:
Tested in PROD:

BEFORE:

Screenshot 2023-07-31 at 15 09 24


AFTER:

Screenshot 2023-07-31 at 15 11 09

Screenshot 2023-07-31 at 15 11 17

graham228221 (approved):

PR #928, opened 2023-09-16
Enforced estimated price

shtukas:
In a previous recent PR ( https://github.com/guardian/price-migration-engine/pull/923 ) we introduced a check that detect when the final price of a subscription after amendment is higher than the estimated price. This is not a condition that would happen naturally in the engine, but it's not a strict impossibility either and it was reported that some Guardian weekly Quarterly (and at least one Annual) subscription had suffered from that problem. 

In this PR 

1. We get rid of `priceCorrectionFactorForPriceCap` which is no longer used. It was introduced as a mechanism for computing the right (possibly capped) prices during the making of charge overrides, but we no longer to that (see 3. below)

2. Replace `PriceCap.cappedPrice` by `PriceCap`'s apply function. This makes the code slightly more readable. 

3. Update the signature of `updateOfRatePlansToCurrent` to no longer passing a price correction factor, but the price we are enforcing. This is a very simple way to solve the mis-pricing problem. Indeed, instead of _hoping_ that the price after amendment will be the one we found during estimation, we simply enforce it. The logic is much more simple to understand and although we keep the price check (just in case), we should no longer observe that kind of mis-pricing. 

4. Note that we are still, at least in the case of legacy migrations, applying the price cap. The three places we are calling `PriceCap` are the notification handler, while making the communication to the user, the Salesforce price rise creation handler, to send the correct (capped) price to Salesforce, and then during the amendment handler. Using the exact same logic during those three handlers, instead of (before this change) using `PriceCap` in the two first but then a correction factor during the last one, was complicating one's mental model of what is actually happening.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -69,18 +69,15 @@ object AmendmentHandler extends CohortHandler {
     }
   }
 
-  private def checkFinalPriceVersusEstimatedNewPrice(
-      cohortSpec: CohortSpec,
-      item: CohortItem,
-      estimatedNewPrice: BigDecimal,
-      newPrice: BigDecimal
-  ): ZIO[Zuora, Failure, Unit] = {
+  private def shouldPerformFinalPriceCheck(cohortSpec: CohortSpec): Boolean = {
     // Date: 8 Sept 2023
     // This function is introduced as part of a multi stage update of the
     // engine to detect and deal with situations where the final price is higher than the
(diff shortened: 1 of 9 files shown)


shtukas:
Subscription 439444, which was the first to alarm after we introduced the final price check: https://github.com/guardian/price-migration-engine/pull/923 is now behaving correctly (meaning relatively to the estimated price, which also happens to be the price cap relatively to its old price)

Before this change:

01 439444 in Zuora with old code

After this change:

03 439444 in Zuora with new code

kelvin-chappell (approved):
🚀

shtukas:
Failure case, obtained by momentarily allowing check for supporter+ and manually editing the cohort item:

Screenshot 2023-09-25 at 19 28 26

PR #939, opened 2023-10-06
Fix code owners

kelvin-chappell:
I think it was broken.

See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners


shtukas (approved):

shtukas:
Merci!

PR #942, opened 2023-10-11
Default mailing address for DigiSubs2023

shtukas:
This PR removes the requirement for a mailing postal address for the (email) notifications of DigiSubs2023. This is a requirement that we inherited from old paper mail migrations but that is needed here, and are causing errors when the user record doesn't have such data in Salesforce. 

We could probably expand that relaxation to all the other digital migrations, but here we focus on the tightest change required to solve that problem with the DigiSubs2023 cohort.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -29,7 +29,7 @@ object NotificationHandler extends CohortHandler {
   // Membership migration
   // Notification period: -33 (included) to -31 (excluded) days
   private val emailMaxNotificationLeadTime = 33
-  private val emailMinNotificationLeadTime = 31
+  private val emailMinNotificationLeadTime = 30
 
   def maxLeadTime(cohortSpec: CohortSpec): Int = {
     MigrationType(cohortSpec) match {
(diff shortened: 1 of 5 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -219,14 +219,24 @@ object NotificationHandler extends CohortHandler {
...
-    } yield billingAddress).orElse(
-      requiredField(contact.MailingAddress, "Contact.MailingAddress")
-    )
+  def targetAddress(
+      cohortSpec: CohortSpec,
+      contact: SalesforceContact
+  ): ZIO[Any, NotificationHandlerFailure, SalesforceAddress] = {
+    MigrationType(cohortSpec) match {
+      case DigiSubs2023 =>
+        ZIO.succeed(SalesforceAddress(Some("street"), Some("city"), Some("state"), Some("postalCode"), Some("country")))
```
kelvin-chappell, line 228: If something has to be passed through, perhaps it would be better if fields are all `None`or could have a `SalesforceAddress.Unused` whose fields are all `None`

shtukas: I tried that, as None was more natural, but that cause additional fields to fail. With that said I guess I can pass `Some("")`, that would also solve the problem and looks more.... canonical :)

shtukas: I did it here: https://github.com/guardian/price-migration-engine/pull/942/commits/499b65a73e81b28aaffc9193bbb2b0da1bd7bac2

kelvin-chappell (approved):
Looks fine 👍 

I'd just say if you want to do a bit of a refactor it's a good practice to keep pure and 'effectful' code separate so that only effectful functions give a `ZIO`.  So where the code is pure, ie doesn't have or consume an effect, it should just return a simpler type and then lift it into a `ZIO` when it gets pulled into code that has an effect.
So, in this case, `targetAddress` might give an `Either[NotificationHandlerFailure, SalesforceAddress]` and then that could be lifted by `ZIO.fromEither` at the call site.

PR #944, opened 2023-10-17
refactor effectful code

shtukas:
In https://github.com/guardian/price-migration-engine/pull/942 , Kelvin was suggesting a small refactoring in the notification handler. In this change, `requiredField`,  `targetAddress` are now both pure functions that return `Either[NotificationHandlerFailure, A]`.


kelvin-chappell (approved):
That's looking much better now! 👍

shtukas:
Flat mapping a left projection, thanks @kelvin-chappell .

PR #948, opened 2023-11-01
DigiSubs2023 Quarterlies and Annuals

shtukas:
This PR implement the tests for the migration of DigiSubs2023 Quarterlies and Annuals. 

Preliminaries
1. https://github.com/guardian/price-migration-engine/pull/949
2. https://github.com/guardian/price-migration-engine/pull/950
3. https://github.com/guardian/price-migration-engine/pull/951
4. https://github.com/guardian/price-migration-engine/pull/952

Cohort specs:

```
{
    "cohortName":"DigiSubs2023_Batch2",
    "brazeCampaignName":"SV_DP_PriceRise2023",
    "importStartDate":"2023-01-01",
    "earliestPriceMigrationStartDate":"2023-12-09"
}
```


lambda/src/test/scala/pricemigrationengine/migrations/DigiSubs2023MigrationTest.scala
```diff
@@ -243,6 +243,245 @@ class DigiSubs2023MigrationTest extends munit.FunSuite {
...
+    assertEquals(newPriceLookup("GBP", Monthly).toOption.get, BigDecimal(14.99))
+    assertEquals(newPriceLookup("GBP", Quarterly).toOption.get, BigDecimal(44.94))
+    assertEquals(newPriceLookup("USD", Annual).toOption.get, BigDecimal(249))
+
+    val account = Fixtures.accountFromJson("DigiSubs2023/quarterly/account.json")
+    val catalogue = Fixtures.productCatalogueFromJson("DigiSubs2023/quarterly/catalogue.json")
+    val subscription = Fixtures.subscriptionFromJson("DigiSubs2023/quarterly/subscription.json")
+    val invoicePreview = Fixtures.invoiceListFromJson("DigiSubs2023/quarterly/invoice-preview.json")
+
+    val zuoraRatePlan1 = subscriptionRatePlan(subscription).toOption.get // If we fail here, that's part of the plan
```
kelvin-chappell, line 263: If you expect this to fail, it would be best to wrap it in an intercept: https://scalameta.org/munit/docs/assertions.html#intercept

kelvin-chappell (approved):
Looks good but would ideally have a single assertion per test so that it's immediately clear what the test is and what needs fixing if it fails.  But no problem with how it is

lambda/src/test/scala/pricemigrationengine/migrations/DigiSubs2023MigrationTest.scala
shtukas: Actually it should never fail. If it does it means that the fixture is broken subscription. Comment is misleading. Let me remove it :)

shtukas:
Yes, I write the test to show how the subscription update is built from parts. In fact it's a unit test of the last assertion with intermediary assertions to highlight the objects :)

PR #955, opened 2023-11-13
Enable nightly DynamoDB backups for CloudFormed tables

jacobwinch:
## What does this change?

This PR allows us to start backing up the `price-migration-engine-cohort-spec` DynamoDB table using https://github.com/guardian/aws-backup. For more details on this project see [this doc](https://docs.google.com/document/d/11EZtuVHCnjavE9AYLroiuDUsMzbp2ulFBtj4VkyCKNc/edit) and this [audit](https://docs.google.com/spreadsheets/d/1gNh_yQ3GVo_dVAemgZfRZM_MaQxep6_x-hzrhvVBGHY/edit#gid=1523470588) that @guardian/devx-reliability have been working on with Heads of Engineering.

## How to test

I should be able to test this by deploying to `CODE`.

I will double check that backups are taken as expected after merging (this should happen the night after this PR is merged).

## How can we measure success?

We will be able to recover (most) data stored in this table in the unlikely event that it is deleted.

## Have we considered potential risks?

Yes, a number of risks related to performance, cost and privacy were considered. See [this doc](https://docs.google.com/document/d/11EZtuVHCnjavE9AYLroiuDUsMzbp2ulFBtj4VkyCKNc/edit#heading=h.vwt7syo8ng40) for more details.

## Question for maintainers of this project

I think the [Riff-Raff project is defined here](https://github.com/guardian/price-migration-engine/blob/03f06fda46c77f63b7f0fb64c55c15fad644f2ce/build.sbt#L36), but it [hasn't been deployed to `CODE` for a long time](https://riffraff.gutools.co.uk/deployment/history?projectName=Retention%3A%3APriceMigrationEngine%3A%3ADynamoDb&stage=CODE&pageSize=20&page=1) and has [never been deployed to `PROD`](https://riffraff.gutools.co.uk/deployment/history?projectName=Retention%3A%3APriceMigrationEngine%3A%3ADynamoDb&stage=PROD&pageSize=20&page=1).

Is deploying via Riff-Raff the correct way to test / release these infrastructure changes?

PR #956, opened 2023-11-15
Stop cloudforming PriceMigrationEngine[Stage]

shtukas:
The PriceMigrationEnginePROD table for the first DynamoDB table created when the engine was born. It is not no longer relevant and can be not only deleted from AWS, but we can stop cloud forming it.

Its existence was noticed as part of DevEx's automatic backup initiative: https://github.com/guardian/price-migration-engine/pull/955 

(Note en passant that the other table declared in that file price-migration-engine-cohort-spec-[Stage] is actively used)


david-pepper (approved):

david-pepper:
typo in description 
``` It is not< no longer relevant ```

kelvin-chappell (approved):
👍

PR #964, opened 2023-12-16
Newspaper2024

shtukas:
Implementation of the Newspaper2024 price migration. We are price rising 3 products according to the below breakdown. 

05 New Prices

This migration is a subscriptions price update (and not a rate plan change). Therefore the Zuora update objects simply apply new prices to the subscriptions bringing them on par with the product catalogue prices that have already been updated in Zuora a few months ago.

This migration is otherwise quite classic, but with a few novelties:

1. We fully carry on with the principle I introduced last year to separate migration specific code into their own namespace. This gives us enormous power in implementing migration specific logic. In fact in this case we have the dedicated Newspaper2024Migration directory with quite a lot of code (and static data) in it.

2. Marketing requested that a small number of subscription be price increase capped at 20%. We did not try and reuse the existing general price capping logic because it would not have work. I actually pointed out how it will be useful to decommission the general price capping functionality that I introduced in late 2022 and let migrations implement their own capping logic if there is one. That general feature is a call for trouble.

3. The testing suite is very thorough, compared to previous migrations, this is due to having to ensure that the code works harmoniously across products.

4. This migration has a non standard notification window relatively to old print products and we used that opportunity to start moving the values to migration specific code (another refactoring that I didn't feel like adding to this PR, will make things more coherent later)

5. Another very non standard feature, was implementing a two weeks pause in the monthlies notifications. This had never been asked by Marketing before and was very very fun to implement, because the engine is not naturally designed to handle non time contiguous notification stages.

6. `SubscriptionData2024` was introduced to make the code manageable more logical and help with testing and. It's been great at those purposes. It's my favourite datatype of this entire project.

7. So.... one very unusual thing with this migration was the fact that for those products, the Zuora product catalogue only has prices for monthly payment variants, but not the Quarterly, Semi Annuals and Annuals payments schedules. For this I introduced the `ChargeDistribution2024` and an algebra to multiply them by a factor. This makes the pricing code very beautiful and predictable and easily testable, and very importantly makes the engine immune to holes in the Zuora product catalogue. Another great effect of that type is that it makes applying capping absolutely trivial.


shtukas:
note: we still need one additional test before I can merge this

lambda/src/main/scala/pricemigrationengine/migrations/DigiSubs2023Migration.scala
```diff
@@ -97,6 +78,7 @@ object DigiSubs2023Migration {
           case None => Left(AmendmentDataFailure(s"Could not determine a new annual price for currency: ${currency}"))
           case Some(price) => Right(price)
         }
+      case SemiAnnual => Left(AmendmentDataFailure(s"There are no defined semi-annual prices fro this migration"))
```
kelvin-chappell, line 81: Small typo:
```suggestion
      case SemiAnnual => Left(AmendmentDataFailure(s"There are no defined semi-annual prices for this migration"))
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationAmendment.scala
```diff
@@ -0,0 +1,89 @@
+package pricemigrationengine.migrations
```
kelvin-chappell, line 1:
```suggestion
package pricemigrationengine.migrations.newspaper2024migration
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationAmendment.scala
```diff
@@ -0,0 +1,89 @@
...
+  def subscriptionToZuoraSubscriptionUpdate(
+      subscription: ZuoraSubscription,
+      effectiveDate: LocalDate,
+  ): Either[AmendmentDataFailure, ZuoraSubscriptionUpdate] = {
+
+    def transform1[T](option: Option[T]): Either[AmendmentDataFailure, T] = {
+      option match {
+        case None                 => Left(AmendmentDataFailure("error"))
+        case Some(ratePlanCharge) => Right(ratePlanCharge)
+      }
```
kelvin-chappell, line 55:
```suggestion
      option.toRight(AmendmentDataFailure("error"))
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationAmendment.scala
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    def transform2[T](data: Either[String, T]): Either[AmendmentDataFailure, T] = {
+      data match {
+        case Left(string) => Left(AmendmentDataFailure(string))
+        case Right(t)     => Right(t)
+      }
```
kelvin-chappell, line 62:
```suggestion
      data.left.map(AmendmentDataFailure)
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+        case Some(ratePlanCharge) => Right(ratePlanCharge)
+      }
+    }
+
+    def transform2[T](data: Either[String, T]): Either[AmendmentDataFailure, T] = {
+      data match {
+        case Left(string) => Left(AmendmentDataFailure(string))
+        case Right(t)     => Right(t)
+      }
+    }
```
kelvin-chappell, line 152: Could extract these

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+        data2024.ratePlanName
+      )
+    } yield price * priceCorrectionFactor
+  }
+
+  def priceData(
+      subscription: ZuoraSubscription,
+  ): Either[AmendmentDataFailure, PriceData] = {
+
+    // PriceData(currency: Currency, oldPrice: BigDecimal, newPrice: BigDecimal, billingPeriod: String)
```
kelvin-chappell, line 138: Not sure this will help

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+
+    case class SubscriptionData2024(
+      productName: String,
+      ratePlan: ZuoraRatePlan,
+      ratePlanName: String,
+      billingPeriod: BillingPeriod,
+      currency: String,
+      currentPrice: BigDecimal
+  )
+   */
```
kelvin-chappell, line 117: Don't think this is helping

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+          currentPrice = currentPrice,
+          targetRatePlanId = targetRatePlanId
+        )) match {
+          case Some(data) => Right(data)
+          case _ => Left(s"[error: 0e218c37] Could not determine SubscriptionData2024 for subscription ${subscription}")
+        }
+      }
+      case _ =>
+        Left(
+          s"[error 93a21a48] Subscription ${subscription.subscriptionNumber} was found to have more than one newsPaperDeliveryRatePlans making determination of rate plan name impossible"
```
kelvin-chappell, line 102: Where does this error code come from?

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+  object MonthliesPart1 extends Newspaper2024BatchId // First batch of monthlies
+  object MonthliesPart2 extends Newspaper2024BatchId // Second batch of monthlies
+  object MoreThanMonthlies extends Newspaper2024BatchId // Quarterlies, Semi-Annuals and Annuals
+
+  def subscriptionToMigrationProductName(subscription: ZuoraSubscription): Either[String, String] = {
+    // We are doing a multi product migration. This function tries and retrieve the correct product given a
+    // subscription.
+    val migrationProductNames = List("Newspaper Delivery", "Newspaper Digital Voucher", "Newspaper Voucher")
+    val names = subscription.ratePlans
+      .filter(ratePlan => ratePlan.lastChangeType == None || ratePlan.lastChangeType == Some("Add"))
```
kelvin-chappell, line 35:
```suggestion
      .filter(ratePlan => ratePlan.lastChangeType.isEmpty || ratePlan.lastChangeType.contains("Add"))
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+      ratePlanCharge <- ratePlan.ratePlanCharges.headOption
+    } yield ratePlanCharge.currency
+  }
+
+  def subscriptionToSubscriptionData2024(subscription: ZuoraSubscription): Either[String, SubscriptionData2024] = {
+    val productName = subscriptionToMigrationProductName(subscription).toOption.get
+    val ratePlans = {
+      subscription.ratePlans
+        .filter(ratePlan => ratePlan.productName == productName)
+        .filter(ratePlan => ratePlan.lastChangeType == None || ratePlan.lastChangeType == Some("Add"))
```
kelvin-chappell, line 70: Could extract this condition as it's repeated above

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+          ratePlan = ratePlan,
+          ratePlanName = ratePlan.ratePlanName,
+          billingPeriod = billingPeriod,
+          currency = currency,
+          currentPrice = currentPrice,
+          targetRatePlanId = targetRatePlanId
+        )) match {
+          case Some(data) => Right(data)
+          case _ => Left(s"[error: 0e218c37] Could not determine SubscriptionData2024 for subscription ${subscription}")
+        }
```
kelvin-chappell, line 98:
```suggestion
        )).toRight(s"[error: 0e218c37] Could not determine SubscriptionData2024 for subscription ${subscription}")
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  def subscriptionToBatchId(subscription: ZuoraSubscription): Either[String, Newspaper2024BatchId] = {
+    val ratePlanDetails = (for {
+      ratePlanDetails <- subscriptionToSubscriptionData2024(subscription)
+    } yield ratePlanDetails)
+
+    ratePlanDetails match {
+      case Left(message) => Left(message)
+      case Right(ratePlanDetails) => {
```
kelvin-chappell, line 173:
```suggestion
    ratePlanDetails.map {
      details => {
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+        ratePlanDetails.billingPeriod match {
+          case Monthly => {
+            val ratePlan = ratePlanDetails.ratePlan
+            ratePlan.ratePlanCharges.toList match {
+              case Nil => Left("")
```
kelvin-chappell, line 178: Assuming `Left` holds an error message to say something's gone wrong, this doesn't look like a useful value.

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
```diff
@@ -0,0 +1,253 @@
...
+
+  def startDateSpreadPeriod(subscription: ZuoraSubscription): Int = {
+    subscriptionToBatchId(subscription) match {
+      case Left(_) => 1
+      case Right(bid) =>
+        bid match {
+          case MonthliesPart1    => 1
+          case MonthliesPart2    => 2
+          case MoreThanMonthlies => 1
+        }
```
kelvin-chappell, line 250: Not sure if it's any clearer but could be:
```suggestion
      case Right(MonthliesPart1)    => 1
          case Right(MonthliesPart2)    => 2
          case Right(MoreThanMonthlies) => 1
        }
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationAmendment.scala
shtukas: That was a great suggestion Kelvin, thanks for that. The individual components can now be more naturally named. This suggests that I should probably do the same for the other migrations, which are sharing the `migrations` namespace. Will do in a separate PR.

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationAmendment.scala
shtukas: Done. Thanks. I didn't know that was a thing. The two `transform1` functions are now gone.

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationAmendment.scala
shtukas: Done. So much better, thank you for that Kelvin. Now the two `transform2` functions are gone. Also updating the PR description

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: Done! Thanks.

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: Nice one. It's now
```
    subscriptionToBatchId(subscription) match {
      case Right(MonthliesPart1)    => 1
      case Right(MonthliesPart2)    => 2
      case Right(MoreThanMonthlies) => 1
      case Left(_)                  => 1
    }
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: Updated to
```
Left(s"Could not extract a rate plan charge for subscription: ${subscription.subscriptionNumber}")
```

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: Here that change would not work as well because the `Right` case can still return a left due to a further check. So going to leave it as a `match` instead of refactoring into a `map` :)

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: The two side by side look like this. 

```
val names = subscription.ratePlans
  .filter(ratePlan => ratePlan.lastChangeType == None || ratePlan.lastChangeType == Some("Add"))
  .map(ratePlan => ratePlan.productName)
  .filter(name => migrationProductNames.contains(name))
  .distinct

val ratePlans = {
  subscription.ratePlans
    .filter(ratePlan => ratePlan.productName == productName)
    .filter(ratePlan => ratePlan.lastChangeType == None || ratePlan.lastChangeType == Some("Add"))
    .distinct
```

I could put `ratePlan.lastChangeType == None || ratePlan.lastChangeType == Some("Add")` into its own function if you feel strongly about it :)

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: On second thought I could still see that those indented checks were not really idiomatic Scala either (no pun intended), and the solution was simply to not exit the `for` construct as I did in the original version. The refactoring that I pushed here ( https://github.com/guardian/price-migration-engine/pull/964/commits/d22540a8046a9ba0967742340c4dcbf84160ee7b ) solves it.

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Newspaper2024MigrationEstimation.scala
shtukas: Never mind. It's actually nicer to abstract it: https://github.com/guardian/price-migration-engine/pull/964/commits/aeac2a052d47ca274bd06827e53d08003eb1aeb8

kelvin-chappell (approved):
Looks great! Nice to see that it can handle so many tricky business requirements.  👍 

I'd say just check that the documentation is keeping up and is sufficient for a new engineer to understand the codebase and what's going on.

And anything you can do to reduce the amount of hardcoded data in the codebase would be good.  If that can be done without an explosion of data sources.

shtukas:
Thank you Kelvin! That one was indeed challenging, but fun. About how the code is evolving, the 2023 migrations and this one in particular have brought the engine outside it's natural operating mode, but I will be able to put more health time into it this year. There are ideas that should be promoted to the common core and things that should be gotten rid of. I have written some documentation already, I will migrate them to the repo :)

lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -97,7 +98,19 @@ object EstimationHandler extends CohortHandler {
 
   def datesMax(date1: LocalDate, date2: LocalDate): LocalDate = if (date1.isBefore(date2)) date2 else date1
```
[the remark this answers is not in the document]
shtukas: I refactored it in its own utility class here: https://github.com/guardian/price-migration-engine/pull/964/commits/924e9f1596888932f91d1e253e3a7a4f63271963 (if there are any other occurrences I will PR them out later)

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Estimation.scala
```diff
@@ -0,0 +1,238 @@
...
+    // - The spread: a mechanism, used for monthlies, by which we do not let a large number of monthlies migrate
+    //   during a single month.
+
+    // We expanded the signature of this function for the Newspaper2024 migration where that date was specific of
+    // the subscription due to its un-unusual scheduling. For Newspaper2024 we call a specific function from the
+    // migration support code.
+
+    val earliestPriceMigrationStartDate = subscriptionToEarliestMigrationStartDate(subscription)
+
+    def datesMax(date1: LocalDate, date2: LocalDate): LocalDate = if (date1.isBefore(date2)) date2 else date1
```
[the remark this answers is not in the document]
shtukas: I refactored it in its own utility class here: https://github.com/guardian/price-migration-engine/pull/964/commits/924e9f1596888932f91d1e253e3a7a4f63271963 (if there are any other occurrences I will PR them out later)

lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -97,7 +98,19 @@ object EstimationHandler extends CohortHandler {
...
 
-  def startDateGeneralLowerbound(cohortSpec: CohortSpec, today: LocalDate): LocalDate = {
+  def startDateGeneralLowerbound(
```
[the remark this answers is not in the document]
shtukas: Thanks. Done

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Estimation.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  // We have an unusual scheduling for this migration and Newspaper2024BatchId is used to
+  // decide the correct start date for each subscription.
+  sealed trait Newspaper2024BatchId
+  object MonthliesPart1 extends Newspaper2024BatchId // First batch of monthlies
+  object MonthliesPart2 extends Newspaper2024BatchId // Second batch of monthlies
+  object MoreThanMonthlies extends Newspaper2024BatchId // Quarterlies, Semi-Annuals and Annuals
+
+  def ratePlanLastChangeTypeIsNoneOrAdd(ratePlan: ZuoraRatePlan): Boolean = {
+    ratePlan.lastChangeType == None || ratePlan.lastChangeType == Some("Add")
```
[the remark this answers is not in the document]
shtukas: Done. Thanks.

lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2024Migration/Estimation.scala
```diff
@@ -0,0 +1,238 @@
...
+    for {
+      ratePlanCharge <- ratePlan.ratePlanCharges.headOption
+      billingPeriod <- ratePlanCharge.billingPeriod
+    } yield BillingPeriod.fromString(billingPeriod)
+  }
+
+  def ratePlanToCurrency(ratePlan: ZuoraRatePlan): Option[String] = {
+    for {
+      ratePlanCharge <- ratePlan.ratePlanCharges.headOption
+    } yield ratePlanCharge.currency
```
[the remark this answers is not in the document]
shtukas: Done!

PR #974, opened 2024-02-07
add originalOrderDate to case class ZuoraRatePlanCharge

shtukas:
Done work in collaboration with @LAKSHMIRPILLAI  and @graham228221 ✨

As part of on going work (refactoring the way the engine decides whether to put a cohort item in cancelled state), we would like a way to attribute an affective date to a rate plan.

The Zuora documentation and our tests show that rate plans themselves to not carry dating information. Moreover the Zuora [documentation](https://developer.zuora.com/api-references/api/operation/GET_SubscriptionsByKey/) is also very unclear about the reponses to the `GET_SubscriptionsByKey` operation and notably the keys of a rate plan charge. In particular they do not mention the key `originalOrderDate` which shows up when we manually query subscriptions. 

To test whether or not we could rely on that key, we tested a large number of subscriptions. We downloaded 1012 subscriptions from Zuora, which yielded 11853 rate plans and 23013 rate plan charges, and 100% of them had a `originalOrderDate` field set.

Considering the above results, the chosen way to attribute a date to a rate plan will be to read the `originalOrderDate` field of any (the first) rate plan charge of the rate plan. (This will be implemented in a coming PR).

This change adds the 

```
originalOrderDate: Option[LocalDate] = None
```

field to the `ZuoraRatePlanCharge` case class.

This caused a lot of tests to have to be updated as we are reading from stored fixtures, causing previously unread `originalOrderDate` to be parsed.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscription.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscription.scala
@@ -54,7 +54,8 @@ case class ZuoraRatePlanCharge(
     billingDay: Option[String] = None,
     triggerEvent: Option[String] = None,
     triggerDate: Option[LocalDate] = None,
-    discountPercentage: Option[Double] = None
+    discountPercentage: Option[Double] = None,
+    originalOrderDate: Option[LocalDate] = None
 )
 
 object ZuoraRatePlanCharge {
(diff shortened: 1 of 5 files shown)


david-pepper (approved):

LAKSHMIRPILLAI (approved):
LGTM 👍🏼

PR #976, opened 2024-02-07
Introduce RateplansProbe and pre Notification subscription check

shtukas:
### Prelude

This change is the follow up of https://github.com/guardian/price-migration-engine/pull/974 and supersede the work done here: https://github.com/guardian/price-migration-engine/pull/925

Work done in collaboration with @LAKSHMIRPILLAI and @graham228221 

### Why

To deal with a pressing problem that presented itself during the digital migrations in 2023, I implemented a test to decide if a subscription should be price rised after its period of inactivity.  There were two problems with it

1. The digital migration were somehow artificially delayed in terms of amendment and the subscriptions that needed to be checked were pre amendment. This was the reason for introducing that test at the beginning of the amendment step but the permanent solution should come at the beginning of the notification step; considering that cohort items sleep pre notification step. 

2. The test was too crude. It was based on a limitation of the Zuora API which only allow the retrieval of the last amendment. This cause(d) too many subscription to be prevented to go through a normal price rise. 

### So what's the thing with the CSRs 

Two things. The set of affected subscriptions was a low ratio but enough for CSRs to report it as something to investigate. Also the timing of the check created confusion to some customers who pointed out that the price rise we have told them would happen didn't happen.

### What's the change ?

The change is that we are completely changing strategy and moving away from the inadequate logic around "last amendment" and instead moving to a much better evaluation that is looking at the entire subscription. More exactly Zuora returns a subscription with its list of expired and active rate plans and we are deciding the result of this check by looking at the array of (active) rate plans. 

For this the following happens in this PR

- We decommission the `IncompatibleAmendmentHistoryFailure` that was being used in the Amendment step. It was introduced just for the original version of that check. It has actually been renamed `RatePlansProbeFailure`.

- Decommission the support functions in the amendment handler: `amendmentIsBeforeInstant` and `checkMigrationRelevanceBasedOnLastAmendment`.

- Make the Zuora service available to the Notification handler. This is to allow that handler to retrieve a subscription as part of processing a cohortItem. Note that this might increase the time taken to process the list of items that need notification and the state machine might need an update. Pascal will keep an eye on that.    

- Introduce `RateplansProbe` which contains the new logic. Also see the comments in the code for guidance. 

- Note that a variety of tests needed to be updated.

- We move the check from the Amendment handler to the Notification handler, where it should have been in the first place, so that if there is a legitimate reason to not price rise a subscription, then that determination will be made before notifying the users.

- Introduce the `RateplansProbeResult` trait, which is the return value of the check. The ternary logic and the fact that one of the outcomes, eg: `IndeterminateConclusion` will cause the engine to fail is on purpose and by design. This will inform us that the probe could not decide and the engine operators will have to identify the type of rate plan that wasn't recognised and update the code. This ensures that we longer generates false positives, but note that the code update will have to be done before the subscription exits its notification window, or otherwise will have to be sent back to `ReadyForEstimation` processing stage. At least, and unlike the current situation, this will have no effect on the users.   
 
 - Added specific tests for `RateplansProbe`.


graham228221 (approved):
Thanks for these changes!

LAKSHMIRPILLAI (approved):

shtukas:
Update: https://github.com/guardian/price-migration-engine/pull/982

PR #979, opened 2024-02-14
add cancellation reason to cohort item

shtukas:
As a follow up of the work we did to redesign the process by which a cohort item can be removed from price rise (meaning put in `Cancelled` processing stage in the engine's Dynamo table), the next step is now to expand the definition of a Cohort Item by adding the optional `cancellationReason` field.

This is a relatively simple process because of the schema-less nature of Dynamo (things would have required a database table definition update if we were using SQL databases)

Although the field is optional, the function `fromCancelledAmendmentResult` is set with a mandatory `reason` field, this will ensure that future cancellations are always given a reason. 

This has also shown that there are now only two reasons why a cohort item can ever be cancelled in the engine: if the corresponding subscription was cancelled in Zuora or if `RateplansProbe.probe` returns the `ShouldCancel` variant of `RateplansProbeResult`, which is exactly what we want.

Note: The files that are generated and sent to the lake have also been extended.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -172,7 +172,14 @@ object NotificationHandler extends CohortHandler {
       case ShouldProceed => ZIO.succeed(())
       case ShouldCancel => {
         val result = CancelledAmendmentResult(item.subscriptionName)
-        ZIO.succeed(CohortTable.update(CohortItem.fromCancelledAmendmentResult(result)).as(result))
+        ZIO.succeed(
+          CohortTable
+            .update(
+              CohortItem
+                .fromCancelledAmendmentResult(result, "(cause: f5c291b0) Subscription was cancelled by RateplansProbe")
+            )
+            .as(result)
+        )
       }
       case IndeterminateConclusion =>
         ZIO.fail(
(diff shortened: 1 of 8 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -172,7 +172,14 @@ object NotificationHandler extends CohortHandler {
       case ShouldProceed => ZIO.succeed(())
       case ShouldCancel => {
         val result = CancelledAmendmentResult(item.subscriptionName)
-        ZIO.succeed(CohortTable.update(CohortItem.fromCancelledAmendmentResult(result)).as(result))
+        ZIO.succeed(
+          CohortTable
+            .update(
+              CohortItem
+                .fromCancelledAmendmentResult(result, "(cause: f5c291b0) Subscription was cancelled by RateplansProbe")
```
graham228221, line 179:
```suggestion
                .fromCancelledAmendmentResult(result, "(cause: f5c291b0) Migration cancelled by RateplansProbe")
```

less ambiguous wording - "subscription" and "cancelled" in a sentence usually has a very specific meaning

shtukas: Love it, thanks! And you are right the word "Subscription" was not the right one here, should have been "Cohort Item" at least :)

graham228221 (approved):
super 😎

kelvin-chappell (approved):
Nice change! 👍 
Now that the semantics of the  `Cancelled` state has changed, which used to mean that the sub had been cancelled and now means that the price rise has been cancelled for that sub, would also be good to update the description here: https://github.com/guardian/price-migration-engine/blob/main/lambda/src/main/scala/pricemigrationengine/model/CohortTableFilter.scala#L32-L36

PR #981, opened 2024-02-15
Upgrade Salesforce API from v43.0 to v60.0

shtukas:
Keeping up to date with Salesforce API numbers, we move from v43.0 to v60.0


LAKSHMIRPILLAI (approved):

graham228221 (approved):
Thanks! 

For reference - there are three new API versions released each year. Salesforce are careful to maintain backwards compatibility, but will occasionally deprecate and retire older API versions. They have recently retired versions <=20.0 and deprecated versions <=30.0.

PR #988, opened 2024-02-23
Refactor PriceCap into LegacyMigrations.priceCap

shtukas:
The PriceCap object was introduced in late 2022 to support a policy of 20% price capping during migrations, but that policy was then abandoned for the 2023 digital migrations, and the Newspaper2024 migration did introduce a price capping, but which was actually incompatible with the original PriceCap and required a custom implementation.

In this PR we refactor PriceCap from `pricemigrationengine.model` to `pricemigrationengine.migrations.LegacyMigrations`. This signals that it's no longer a primary concept of the engine and kept to support the legacy migrations. Moreover any future migration price caps (complete or partial, with fixed or varying percentage) should be seen as specific to migrations and implemented as such.


lambda/src/test/scala/pricemigrationengine/handlers/NotificationHandlerTest.scala
```diff
@@ -28,16 +29,15 @@ class NotificationHandlerTest extends munit.FunSuite {
...
   private val estimatedNewPrice = BigDecimal(15.00)
-  test("For membership test, we need the estimatedNewPrice to be higher than the capped price") {
-    assert(PriceCap(oldPrice, estimatedNewPrice) < estimatedNewPrice)
+  test("For legacy migrations, we need the estimatedNewPrice to be higher than the capped price") {
+    assert(LegacyMigrations.priceCap(oldPrice, estimatedNewPrice) < estimatedNewPrice)
   }
 
-  // The price that is displayed to the customer is capped using the old price as base
-  private val cappedEstimatedNewPriceWithCurrencySymbolPrefix = "£12.00"
+  private val estimatedNewPriceWithCurrencySymbolPrefix = "£15.0"
```
graham228221, line 36: should this be:

```suggestion
  private val estimatedNewPriceWithCurrencySymbolPrefix = "£15.00"
```

not sure how precise this text value needs to be.

graham228221 (approved):

lambda/src/test/scala/pricemigrationengine/handlers/NotificationHandlerTest.scala
shtukas: Good observation. The string that is being tested against is the value that is going to be printed into the email or the letter. When the price is essentially a whole number it is cast to a float, and for some reasons printed with one zero, but in the case the price was capped and the new capped price also an integer (which happens when the old price was `10`, the new price should have been `15`, but the 20% capped price is `12`), then it is printed with 2 zeros. 

It looks like the capping calculation changes _something_ in the way the number is going to be printed 🙃

I don't think it's a problem, the email and letters have been visually reviewed, but it's an interesting observation :)

PR #989, opened 2024-02-26
Run notification in batches

shtukas:
The ongoing Newspaper2024 migrates a large number of subscriptions. This, compounded with the change in the notification handler we implemented recently ( https://github.com/guardian/price-migration-engine/pull/976 ), causes the lambda to overrun the 15 mins limit imposed by AWS. Here, we upgrade the state machine and the Notification handler to run on small batches, following the same pattern we are already using for the Estimation and Amendment handlers.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -13,7 +13,6 @@ import pricemigrationengine.migrations.newspaper2024Migration
   */
 object AmendmentHandler extends CohortHandler {
 
-  // TODO: move to config
   private val batchSize = 50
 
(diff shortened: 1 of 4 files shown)


LAKSHMIRPILLAI (approved):

shtukas:
04 Screenshot 2024-02-26 at 12 28 28

kelvin-chappell (approved):
Looks good! 👍 

One thing I discovered recently is that the [lambda context](https://docs.aws.amazon.com/lambda/latest/dg/java-context.html) has a `getRemainingTimeInMillis` property so it would be possible in theory to work out if you're going to have time to process another sub by the time it's taken you to process the subs you've already done.  Then you wouldn't need a hardcoded batch size.  But it would add complexity and you'd have to work out a reasonable time to leave for the next one.  So almost certainly not worth doing.

PR #1001, opened 2024-03-07
Move the start date computation support function to a utility object

shtukas:
The computation of the start date (part of the estimation step), now relies on a larger set of support functions required to implement various policies and migration specific requirements. This change aggregates them into a separate new utility object called `StartDates`, for simplicity and increase code understanding.


lambda/src/main/scala/pricemigrationengine/util/StartDates.scala
```diff
@@ -0,0 +1,106 @@
...
+  amendment are taking effect.
+ */
+
+object StartDates {
+
+  def cohortSpecLowerBound(
+      cohortSpec: CohortSpec,
+      today: LocalDate
+  ): LocalDate = {
+    // The startDateGeneralLowerBound is a function of the cohort spec and the notification min time.
```
LAKSHMIRPILLAI, line 32: startDateGeneralLowerBound in the comment must be cohortSpecLowerBound(?

LAKSHMIRPILLAI (approved):

PR #1013, opened 2024-03-15
ZuoraRatePlan: enforce naming convention

shtukas:
We enforce the naming convention that is being used with the other functions on the ZuoraRatePlan object.


graham228221 (approved):
I like this little change!

PR #1019, opened 2024-03-19
GW2024 Part 4: Estimation Step

shtukas:
Previously:
- Part 1: https://github.com/guardian/price-migration-engine/pull/1012
- Part 2: https://github.com/guardian/price-migration-engine/pull/1016
- Part 3: https://github.com/guardian/price-migration-engine/pull/1018

This is the fourth PR in our series where we are using the GW2024 migration to show how to organise the changes required to build a modern migration.

Here we prepare the Estimation steps. For this we do the following

1. The core of the GW2024 Estimation step is the `GW2024.priceData` function. It's the GW2024 implementation of the more general `AmendmentData.priceData`. The set of functions we added to `GW2024Migration` is the smallest set required for this function to return a value.

2. Added tests to cover every function we added in (1.)

3. Plugged in `StartDates.lastPriceRiseDate` the GW2024 implementation of that function.

4. We opened the Estimation Lambda by removing the lock we had against cohort name "GW2024"

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
@@ -103,18 +103,14 @@ object EstimationHandler extends CohortHandler {
   }
 
   def handle(input: CohortSpec): ZIO[Logging, Failure, HandlerOutput] = {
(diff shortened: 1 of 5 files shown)


tomrf1 (approved):
👍

PR #1026, opened 2024-03-26
Better determination of migration rate plans for GW2024

shtukas:
This part of the ongoing series of PRs to show how a migration is implemented using the case of GW2024 as example. 

Here we are not adding a new step yet, but making an improvement to the code of the Estimation Step ( https://github.com/guardian/price-migration-engine/pull/1019 ). A better way to determine the active rate plan. And notably notes on dealing with the fact that Zuora doesn't always provide the `lastChangeType` attribute of a rate plan.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/GW2024Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/GW2024Migration.scala
@@ -39,6 +39,30 @@ object GW2024Migration {
     "ROW (USD)" -> BigDecimal(396)
   )
 
+  /*
+
+      Date: 27 March 2024
+      Author: Pascal
+
+      I have downloaded all the Zuora subscriptions using the subscription
+      numbers from the migration cohort ids and extracted the rate plan names. Found:
+
+      GW Oct 18 - Monthly - Domestic
+      GW Oct 18 - Quarterly - Domestic
+      GW Oct 18 - Annual - Domestic
+      GW Oct 18 - Monthly - ROW
+      GW Oct 18 - Quarterly - ROW
+      GW Oct 18 - Annual - ROW
+      Guardian Weekly Quarterly
+      Guardian Weekly Annual
+      Guardian Weekly 1 Year
+
+      It is not necessary to do this as part of preparing and writing the code for a migration,
+      but doing so gives us an exact list of the rate plan names that we are looking for
+      and this simplifies the extraction of the rate plan that is going to be upgraded for
+      each subscription.
+   */
+
   val migrationRatePlanNames: List[String] = List(
     "GW Oct 18 - Monthly - Domestic",
     "GW Oct 18 - Quarterly - Domestic",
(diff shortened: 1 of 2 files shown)


lambda/src/main/scala/pricemigrationengine/migrations/GW2024Migration.scala
```diff
@@ -48,33 +64,35 @@ object GW2024Migration {
...
-    )
-    subscription.ratePlans.filter(rp => migrationRatePlanNames.contains(rp.ratePlanName))
+    subscription.ratePlans
+      .filter(rp => lastChangeTypeIsNotRemove(rp))
+      .filter(rp => migrationRatePlanNames.contains(rp.ratePlanName))
   }
 
   def subscriptionToMigrationRatePlan(subscription: ZuoraSubscription): Option[ZuoraRatePlan] = {
-    subscriptionToMigrationRatePlans(subscription: ZuoraSubscription).headOption
+    subscriptionToMigrationRatePlans(subscription) match {
```
tomrf1, line 88: what's the reason for replacing `headOption` here?

lambda/src/main/scala/pricemigrationengine/migrations/GW2024Migration.scala
```diff
@@ -48,33 +64,35 @@ object GW2024Migration {
...
-  // Data Extraction and Manipulation
-  // ------------------------------------------------
+  // Note about the difference between lastChangeTypeIsAdd and lastChangeTypeIsNotRemove
+  // Sadly `lastChangeType` is not always defined on all rate plans. The situation is:
+  //     - Not defined                    -> Active rate plan
+  //     - Defined and value is "Add"     -> Active rate plan
+  //     - Defined and value is "Removed" -> Non active rate plan
+
+  def lastChangeTypeIsAdd(ratePlan: ZuoraRatePlan): Boolean = {
+    ratePlan.lastChangeType.fold(false)(_ == "Add")
```
kelvin-chappell, line 74:
```suggestion
    ratePlan.lastChangeType.contains("Add")
```

lambda/src/main/scala/pricemigrationengine/migrations/GW2024Migration.scala
```diff
@@ -48,33 +64,35 @@ object GW2024Migration {
...
+  }
+
+  def lastChangeTypeIsNotRemove(ratePlan: ZuoraRatePlan): Boolean = {
+    ratePlan.lastChangeType.fold(true)(_ != "Remove")
```
kelvin-chappell, line 78:
```suggestion
    !ratePlan.lastChangeType.contains("Remove")
```

lambda/src/main/scala/pricemigrationengine/migrations/GW2024Migration.scala
shtukas: @tomrf1 

The code that you asked the question about is a previous version. I have updated the implementation in subsequent commits.  

That having been said, let me answer your question because it's an interesting one. It was a precaution. I wanted to take account of the extremely unlikely but still not impossible situation of a subscription having two rate plans of different names, both names appearing in the list of migration rate plan names and both being active. Something like that would not happen in practice, but could happen in theory. 

In any case, the above is now completely irrelevant, because I downloaded all 33,000+ subscriptions of the migration and tested all of them one by one. They are fine, so a lazier implementation (eg: going back to `headOption`) works perfectly fine.

shtukas:
Thanks for the point of syntax @kelvin-chappell :)

tomrf1 (approved):

PR #1030, opened 2024-04-04
Refactoring: Update the Fixtures Directory

shtukas:
In this refactoring we update the structure of the `test/resources` directory to mirror the conventions of separating migration related tests into their own sub-directories. This will facilitate future cleaning/decommissions and also clarify the difference between core tests and migration related tests.


tomrf1 (approved):
👍 makes sense

PR #1038, opened 2024-04-23
handle amendment failed due to lock

shtukas:
6 days ago we made this PR: https://github.com/guardian/price-migration-engine/pull/1035 to prevent a rare situation where Zuora would perform two price rises on a subscription. At this aim we succeeded (as far as Pascal can tell), but this created another problem...

Each day the DigiSubs2023_Batch2 migration would fail. More exactly the state machine would be interrupted with the error: "ZuoraUpdateFailure: Operation failed due to a lock competition, please retry later". Interestingly just rerunning the Amendment lambda would fix it, without any particular correction needed to be applied in Zuora or on the DynamoDB cohort item. The only problem is that it would then happen a dozen times across all amendments for the day. 

Originally we thought it was limited to DigiSubs2023_Batch2, but a couple of days ago it also started affecting GW2024 too.

We did not want to reintroduce the blanket automatic retry (for obvious reasons), the solution was to capture the ZuoraUpdateFailures corresponding to a "lock competition", and return a `ZIO.succeed` without updating the cohort item. 

We have just seen it happening in PROD 🎉

Screenshot 2024-04-23 at 23 51 26

When an item is being locked for amendment by Zuora, it will simply be amended the next time the lambda runs. This may be on the same day, or the day after (if it was the last run of the amendment for that day).

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -52,6 +52,16 @@ object AmendmentHandler extends CohortHandler {
           val result = ExpiringSubscriptionResult(item.subscriptionName)
           CohortTable.update(CohortItem.fromExpiringSubscriptionResult(result)).as(result)
         }
+        case e: ZuoraUpdateFailure => {
+          // We are only interested in the ZuoraUpdateFailures corresponding to message
+          // "Operation failed due to a lock competition"
+          // We succeed them without cohort item update to be done later.
+          if (e.reason.contains("lock competition")) {
+            ZIO.succeed(AmendmentPreventedDueToLockResult(subscriptionNumber = item.subscriptionName))
+          } else {
+            ZIO.fail(e)
+          }
+        }
         case e => ZIO.fail(e)
       },
       success = { result =>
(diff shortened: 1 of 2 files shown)


shtukas:
@kelvin-chappell

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -52,6 +52,16 @@ object AmendmentHandler extends CohortHandler {
           val result = ExpiringSubscriptionResult(item.subscriptionName)
           CohortTable.update(CohortItem.fromExpiringSubscriptionResult(result)).as(result)
         }
+        case e: ZuoraUpdateFailure => {
+          // We are only interested in the ZuoraUpdateFailures corresponding to message
+          // "Operation failed due to a lock competition"
+          // We succeed them without cohort item update to be done later.
+          if (e.reason.contains("lock competition")) {
```
kelvin-chappell, line 59: The reason field is very unlikely to change but if there's anything like a message ID that might be more reliable.

kelvin-chappell (approved):
👍

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
shtukas: The error code is "53500050". Zuora doesn't have an entry for that one, but the documentation says that the way error codes are built, the last two digits are relevant and "50" occurs when a _Locking contention_.

_This request cannot be processed because the objects this request is trying to modify are being modified by another API request, UI operation, or batch job process._

I could switch to search for the entire code inside the string but that feels like as arbitrary as the pattern. Going to keep the pattern for the moment 🙂

PR #1045, opened 2024-05-16
move aws lambda runtime to java17

shtukas:
Move the aws lambda runtime from java8.al2 to java11

An updated version of ZIO, which was introduced here: https://github.com/guardian/price-migration-engine/pull/1042, is not compatible with java8

Note that we originally attempted to migrate to Java 21 or Java 17 but encountered the problem described here: https://github.com/guardian/support-service-lambdas/pull/2204, so we are just migrating to 11 for the moment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         uses: actions/setup-java@v4.2.1
         with:
           distribution: corretto
-          java-version: 17
+          java-version: 11
           cache: sbt
 
       - name: Test PR from forked repo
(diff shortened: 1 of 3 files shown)


kelvin-chappell:
Why not move to Java 21? 
https://github.com/guardian/recommendations/blob/7b023a0d26482ae34e3c5644aea1b2a3104562d3/scala.md?plain=1#L9
It's the [latest LTS version](https://aws.amazon.com/corretto/faqs/)

shtukas:
Merci @kelvin-chappell Done. I didn't immediately see 21 was an option in the AWS documentation. Cheers!

jamesgorrie (approved):

kelvin-chappell (approved):
👍

shtukas:
@kelvin-chappell @jamesgorrie 
FYI, I just came across the same issue as https://github.com/guardian/support-service-lambdas/pull/2204
Looks like we can't have 21 and that I need to downgrade to 17 or 11.

PR #1047, opened 2024-05-21
Notify Salesforce of migration interrupted at Notification with a reason

shtukas:
# Current Situation

The engine notifies Salesforce of the progress of a subscription migration in three places.

1. The SalesforcePriceRiseCreationHandler where we notify Salesforce that a subscription has been scheduled for migration (this happens just after the estimation step).

2. The SalesforceNotificationDateUpdateHandler, where we notify Salesforce that a subscription has just undergone its notification step.

3. The SalesforceAmendmentUpdateHandler, where we notify Salesforce that a subscription has been amended in Zuora.

When undergoing the estimation step, the engine may discover that the subscription was cancelled in Zuora. In this case it it marked as cancelled in the DynamoDB and Salesforce is not notified of anything.

When undergoing the notification step, the engine can discover that the subscription had been cancelled in Zuora, in this case the subscription is marked as cancelled in the DynamoDB. The engine can also discover, thanks to the work done in RateplansProbe that the subscription is in a state incompatible with the price rise, in this case the event is marked into the DynamoDB.

# This change

Here we add to the logic of the engine a notification sent to Salesforce to inform Salesforce of the case of a migration which has been interrupted at notification step because the subscription was found to have been cancelled in Zuora, but also the case where the subscription fails its RateplansProbe check.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -373,23 +373,6 @@ object NotificationHandler extends CohortHandler {
     } yield ()
   }
 
-  private def buildPriceRiseForCancellation(
-      cohortSpec: CohortSpec,
-      cohortItem: CohortItem,
-      reason: Option[String]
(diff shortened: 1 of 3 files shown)


graham228221 (approved):
Great, thanks for this!

PR #1056, opened 2024-07-10
SupporterPlus 2024

shtukas:
Here we implement the SupporterPlus 2024 price migration. 

This migration is a standard migration in terms of updating subscriptions' prices within the same rate plan, but important differences with more regular migrations (for instance GW2024 and Newspaper2024 we did earlier this year) are: 

1. We have a new processing stage called `DoNotProcessUntil`, which comes with an extension of `CohortItem`. A new optional attribute that carries a LocalDate. Items are put into that stage when, for any general reason, we want the item to sleep for a period of time before being estimated (or re-estimated). 

2. The first use of this new processing stage is to put cohort items to sleep if at notification stage we notice an active cancellation discount on them. Where "active" means that it was issued less than six months ago, in which case we make the item sleep until the 6th month anniversary of the effective date of the discount. 

3. We price rise only one of the two charges of the rate plan, and leave the other one alone. SupporterPlus subscriptions have a base charge and an optional contribution charge. The price rise has been applied to the base charge, leaving the existing contribution unchanged. Moreover we cap the rise at 27%. This price rise of essentially a part of a rate plan, has caused some complexity in the estimation handler and the Notification handler, but notably the salesforce price rise notification handler (which runs just after the Estimation), has ran once on an ad-hoc script and will need to be updated within the next six months, when the first items that received a cancellation save discount in Zuora, and are currently in DoNotProcessUntil processing stage in the DynamoDB will need to be re-estimated.

4. For the user notification, we used two Braze emails templates, which were different depending on whether or not the subscription had a non zero contribution. This was totally novel and required special adjustments of the Notification handler for this migration. As a result the `CohortSpec`'s `brazeCampaignName` was not used for this migration and the names of the two canvases hardcoded. Depending on whether marketing develops a taste for this kind of feature, we may end up moving the `brazeCampaignName` out of the `CohortSpec`.

5. In addition of using two Braze canvas templates, we also extended the user custom attributes of the message sent to braze, to include the breakdown of the base price changes and contribution amounts for better description of the price rise to the users. This has causes ad-hoc changes in the notification handler that are completely specific to SupporterPlus2024. This in fact broke the principle of concentrating in the migration objects ideas the idiosyncrasies of a specific migration. Then this migration ends, we will remove those changes from the code (they are clearly indicated as being specific to SupporterPlus 2024).


lambda/src/main/scala/pricemigrationengine/handlers/EstimationHandler.scala
```diff
@@ -32,6 +32,36 @@ object EstimationHandler extends CohortHandler {
...
+        } else { ZIO.succeed(()) }
+    } yield ()
+  }
+
+  def monitorDoNotProcessUntils(today: LocalDate): ZIO[Logging with CohortTable, Failure, Unit] = {
+    // This function migrates items in DoNotProcessUntil state, which have passed their
+    // expiration time, to ReadyForEstimation
+    for {
+      _ <- CohortTable
+        .fetch(DoNotProcessUntil, None)
```
tomrf1, line 55: is there a date filter for this fetch call?

shtukas: @tomrf1 

Good question. Answer is No. 

That function is defined here: https://github.com/guardian/price-migration-engine/blob/db76ee8627aa1b7732f3479c92c7a3f9792ddff8/lambda/src/main/scala/pricemigrationengine/services/CohortTableLive.scala#L127 and as you can see it either (if the second argument is `None`) retrieves all the items of a given processing stage or (if the second argument is defined) all items of a given processing stage whose `startDate` is before a given date. 

Most times it's used, for instance here https://github.com/guardian/price-migration-engine/blob/db76ee8627aa1b7732f3479c92c7a3f9792ddff8/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala#L22 , we do not specify a date because we want all the items. 

In this case I want all the items in `DoNotProcessUntil`, and this function https://github.com/guardian/price-migration-engine/pull/1056/files#diff-23bf94a527ab471ddb3378219c90b6bc237922cbc47e53ea002e247b611a5e6eR35 will decide what to do with individual items.

docs/coding-directives.md
```diff
@@ -0,0 +1,8 @@
+# Coding Conventions
```
jamesgorrie, line 1:
```suggestion
# Coding Directives
```

lambda/src/main/scala/pricemigrationengine/model/membershipworkflow/EmailMessage.scala
```diff
@@ -18,9 +18,60 @@ case class EmailPayloadSubscriberAttributes(
     next_payment_date: String,
     payment_frequency: String,
     subscription_id: String,
-    product_type: String
+    product_type: String,
+
+    // SupporterPlus 2024 extension (see comment below)
```
tomrf1, line 23: why not make these Options?

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -113,10 +148,27 @@ object NotificationHandler extends CohortHandler {
...
         case GW2024 =>
           s"${currencySymbol}${PriceCap.priceCapForNotification(oldPrice, estimatedNewPrice, GW2024Migration.priceCap)}"
+        case SupporterPlus2024 => s"${currencySymbol}${estimatedNewPrice}"
       }
 
       _ <- logMissingEmailAddress(cohortItem, contact)
 
+      // ----------------------------------------------------
+      // Data for SupporterPlus2024
+      subscription <- Zuora.fetchSubscription(cohortItem.subscriptionName)
```
tomrf1, line 158: do we only need to call out to zuora if it's SupporterPlus?

lambda/src/main/scala/pricemigrationengine/model/BillingPeriod.scala
```diff
@@ -13,11 +13,11 @@ object BillingPeriod {
...
-    "Month" -> "Monthly",
-    "Quarter" -> "Quarterly",
-    "Quarterly" -> "Quarterly",
-    "Semi_Annual" -> "Semiannually",
-    "Annual" -> "Annually"
+    "Month" -> "monthly",
+    "Quarter" -> "quarterly",
+    "Quarterly" -> "quarterly",
+    "Semi_Annual" -> "semiannually",
+    "Annual" -> "annually"
```
tomrf1, line 20: what's the reason for changing the format here?

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
shtukas: @tomrf1 
Oh. This is actually a very good point. I should refactor this. I will make it so that a separate function returns the data that is being computed in that section and that function will only call Zuora for SupporterPlus2024. Good suggestion!

lambda/src/main/scala/pricemigrationengine/model/BillingPeriod.scala
shtukas: @tomrf1 
Request from marketing. The change will affect all the migrations, but it was always wrong to start with. It means that all emails and letter will have a better capitalization from now on.

lambda/src/main/scala/pricemigrationengine/model/membershipworkflow/EmailMessage.scala
shtukas: You mean this?

```
 sp2024_contribution_amount: String = ""
 sp2024_previous_combined_amount: String = ""
 sp2024_new_combined_amount: String = ""
```

That's a good question. I tried that, but then wanted to remain backward compatible with existing tests which would have needed to be updated, and I didn't want to do that. Considering that this is code that is going to be thrown away in 13 months, I didn't think of it as a big problem.

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
shtukas: @tomrf1 
Refactoring done: https://github.com/guardian/price-migration-engine/pull/1056/commits/1349068f26c7748c40c674408914201696386f10

lambda/src/main/scala/pricemigrationengine/model/membershipworkflow/EmailMessage.scala
shtukas: @tomrf1 
Actually wasn't too bad. https://github.com/guardian/price-migration-engine/pull/1056/commits/f66073fc1997309fb0d1e7960502934ac963fb5b

tomrf1 (approved):
looks good 👍

PR #1085, opened 2024-10-21
Orders Harmonization: Subscription Renewal

shtukas:


shtukas:
Tested in PROD. This renewal was using the new API:

02 Screenshot 2024-10-21 at 15 24 28

tomrf1 (approved):
👍

shtukas:
Just made the interesting discovery that the Orders API fail with a 200, but has a answer that can be introspected, for instance

```
{
  "success" : false,
  "processId" : "A6F3402C9882F60F",
  "reasons" : [ {
    "code" : 58560030,
    "message" : "The effective date of the operation (2024-11-24) should not be later than the term end date of the basic subscription (2024-11-08)."
  } ],
  "requestId" : "9b458b38-d731-416a-ad69-1272d8e4794f"
}
```

shtukas:
The answer in case of successful processing has the following shape

```
{
  "success" : true,
  "orderNumber" : "O-03573572",
  "accountNumber" : "A0234567",
  "status" : "Completed",
  "subscriptionNumbers" : [ "A-S0198765" ]
}
```

PR #1087, opened 2024-10-24
Orders Harmonization: Subscription Amendment 

shtukas:
This PR is the second part of the migration to the new Zuora Orders API. ( In part 1 we have updated the subscription renewal function: https://github.com/guardian/price-migration-engine/pull/1085 )

In this part we cannot just update the ZuoraLive function that is performing the update, mostly because of signature differences between the old API and the new one that cause changes in the way the amendment handler is structured. Consequently we are performing that migration only for SupporterPlus2024, while leaving DigiSubs2023_Batch2, Newspaper2024 and GW2024 untouched.

The situation with the engine is now 

```
subscription renew: new Orders API for all migrations

amendment:
  DigiSubs2023_Batch2  : old API
  Newspaper2024        : old API
  GW2024               : old API
  SupporterPlus2024    : new Orders API
```

Although the three first migrations are still using the old API, we will not specifically migrate them to the new API, because they will be completed by the time the old API is decommissioned. Another reason for not migrating them is that although the migration of SupporterPlus2024 makes the engine de facto support the new API, slight diffrences between the ways the Order is to be build between migrations is not something that it necessarily worth doing.

The model hierarchy presented in this PR captures the making of a ZuoraAmendmentOrderPayload according
to the schema presented here: https://knowledgecenter.zuora.com/Zuora_Billing/Manage_subscription_transactions/Orders/Order_actions_tutorials/D_Replace_a_product_in_a_subscription

One interesting note about the design is using a sealed trait to represent the different types of an Order action,
in the context of replacing a product in a subscription. Indeed, the two actions in that order are:

```
{
  "type": "RemoveProduct",
  "triggerDates": [
      {
          "name": "ContractEffective",
          "triggerDate": "2024-11-28"
      },
      {
          "name": "ServiceActivation",
          "triggerDate": "2024-11-28"
      },
      {
          "name": "CustomerAcceptance",
          "triggerDate": "2024-11-28"
      }
  ],
  "removeProduct": {
      "ratePlanId": "8a12867e92c341870192c7c46bdb47d6"
  }
}
```

and

```
{
  "type": "AddProduct",
  "triggerDates": [
      {
          "name": "ContractEffective",
          "triggerDate": "2024-11-28"
      },
      {
          "name": "ServiceActivation",
          "triggerDate": "2024-11-28"
      },
      {
          "name": "CustomerAcceptance",
          "triggerDate": "2024-11-28"
      }
  ],
  "addProduct": {
      "productRatePlanId": "8a128ed885fc6ded018602296ace3eb8",
      "chargeOverrides": [
          {
              "productRatePlanChargeId": "8a128ed885fc6ded018602296af13eba",
              "pricing": {
                  "recurringFlatFee": {
                      "listPrice": 12
                  }
              }
          },
          {
              "productRatePlanChargeId": "8a128d7085fc6dec01860234cd075270",
              "pricing": {
                  "recurringFlatFee": {
                      "listPrice": 0
                  }
              }
          }
      ]
  }
}
```

They are respectively modeled as `ZuoraAmendmentOrderPayloadOrderActionRemove` and `ZuoraAmendmentOrderPayloadOrderActionAdd`

Because of the way upickle works, we end up with a JSON serialization that has a "$type" field which is a bit annoying. We can see it in the test data in SupporterPlus2024MigrationTest.

I have tried to submit the Orders payload to Zuora with the "$type" field left inside the JSON but Zuora errors during the processing. (Technically, it should actually work, but I think that Zuora may be doing payload introspection beyond the mandatory fields and is error'ing on the "$type" field)

I have tried to remove the "$type" field by providing a custom writer for the serialization of the two classes, but could not get that to work.

In the end the solution I have chosen was to remove the "$type" field from the JSON before sending it to Zuora, which explains the function type_flush in the ZuoraLive implementation of applyAmendmentOrder

This is, for all intent and purpose a hack and in a future change we may try and get those custom writers to work to avoid string manipulation in the implementation of applyAmendmentOrder.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/services/ZuoraLive.scala
+++ b/lambda/src/main/scala/pricemigrationengine/services/ZuoraLive.scala
@@ -208,11 +208,37 @@ object ZuoraLive {
           )
         }
 
+        override def applyAmendmentOrder(
+            subscription: ZuoraSubscription,
+            payload: ZuoraAmendmentOrderPayload
+        ): ZIO[Any, ZuoraOrderFailure, Unit] = {
+          post[ZuoraAmendmentOrderResponse](
+            path = s"orders",
+            body = write(payload)
+          ).foldZIO(
+            failure = e =>
+              ZIO.fail(
+                ZuoraOrderFailure(
+                  s"[f8569839] subscription number: ${subscription.subscriptionNumber}, payload: ${payload}, reason: ${e.reason}"
+                )
+              ),
+            success = response =>
+              if (response.success) {
+                ZIO.succeed(())
+              } else {
+                ZIO.fail(
+                  ZuoraOrderFailure(
+                    s"[bb6f22ef] subscription number: ${subscription.subscriptionNumber}, payload: ${payload}, with answer ${response}"
+                  )
+                )
+              }
+          )
+        }
+
         override def renewSubscription(
             subscriptionNumber: String,
             payload: ZuoraRenewOrderPayload
         ): ZIO[Any, ZuoraRenewalFailure, Unit] = {
-
           post[ZuoraRenewOrderResponse](
             path = s"orders",
             body = write(payload)
@@ -220,7 +246,7 @@ object ZuoraLive {
             failure = e =>
               ZIO.fail(
                 ZuoraRenewalFailure(
-                  s"[06f5bd6f] subscription number: $subscriptionNumber, payload: ${payload}, reason: ${e.reason}"
+                  s"[06f5bd6f] subscription number: ${subscriptionNumber}, payload: ${payload}, reason: ${e.reason}"
                 )
               ),
             success = response =>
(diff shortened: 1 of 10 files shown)


shtukas:
Tested: The two following amendments have been made, by two orders, one for the renewal and one for the rate plan upgrade:

Screenshot 2024-10-26 at 22 49 03

lambda/src/main/scala/pricemigrationengine/services/ZuoraLive.scala
```diff
@@ -208,11 +208,51 @@ object ZuoraLive {
...
+
+          // The existence of type_flush is explained in comment cada56ad.
+          // This is, for all intent and purpose a hack due to the way upickle deals with sealed traits
+          // and in a future change we will get rid of it, either by changing JSON library or coming up
+          // with the correct writers.
+
+          def type_flush(str: String): String = {
+            str
+              .replace(""""$type":"ZuoraAmendmentOrderPayloadOrderActionAdd",""", "")
+              .replace(""""$type":"ZuoraAmendmentOrderPayloadOrderActionRemove",""", "")
```
graham228221, line 224: Don't worry, we have to do this in Salesforce all the time as Apex has a lot of reserved keywords like `type`, `currency`, etc that causes issues with json parsing 😁

graham228221 (approved):
This looks great, thanks for walking me through this.

PR #1089, opened 2024-10-25
Add id to ZuoraSubscription

shtukas:
Add the `id` attribute to a ZuoraSubscription which, interestingly, didn't have it. (This is preliminary work for the Zuora Orders migration for amendments).

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscription.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/ZuoraSubscription.scala
@@ -7,6 +7,7 @@ import upickle.default._
 
 case class ZuoraSubscription(
     subscriptionNumber: String,
+    id: String,
     version: Int,
     customerAcceptanceDate: LocalDate,
     contractEffectiveDate: LocalDate,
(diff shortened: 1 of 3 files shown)


graham228221 (approved):

LAKSHMIRPILLAI (approved):

shtukas:
Motivation: 

The new subscription Id (post amendment), is written in the DynamoTable here: https://github.com/guardian/price-migration-engine/blob/86c3a806023008d3fb186947f8ad24aeadb20221/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala#L213 (more exactly is used to make an instance of `SuccessfulAmendmentResult` which is then written to the table.)

We get it in the answer of the amendment call itself: https://github.com/guardian/price-migration-engine/blob/86c3a806023008d3fb186947f8ad24aeadb20221/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala#L179

But since I will be using the new API, I would like to be able to query Zuora post calling the new API and read it from the subscription itself.

PR #1099, opened 2024-11-28
compute the doNotProcessUntil timespan for cancellation saves

shtukas:
This updates the logic of the interaction between subscriptions in "cancellation save" status and a pending price migration, a feature that we added for Supporter+ 2024.

In this update the time that the item is going to wait before being re-estimated, is computed as a function of the billing period. In essence we want to wait (at least) two billing periods before notifying the user of a price migration, instead of 6 months.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -79,16 +79,30 @@ object NotificationHandler extends CohortHandler {
       item: CohortItem,
       subscription: ZuoraSubscription
   ): ZIO[CohortTable with Zuora, Failure, Unit] = {
+    // Known keys for this mapping can be found in the BillingPeriod object
+    val mapping: Map[String, Int] = Map(
+      "Month" -> 1,
+      "Quarter" -> 4,
+      "Quarterly" -> 4,
+      "Semi_Annual" -> 6,
+      "Annual" -> 12
+    )
     for {
       cancellationDate <- ZIO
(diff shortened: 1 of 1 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -79,16 +79,30 @@ object NotificationHandler extends CohortHandler {
       item: CohortItem,
       subscription: ZuoraSubscription
   ): ZIO[CohortTable with Zuora, Failure, Unit] = {
+    // Known keys for this mapping can be found in the BillingPeriod object
```
tomrf1, line 82: not very important, but a more robust approach is to use `BillingPeriod.fromString` and have an exhaustive pattern match instead of a Map

shtukas: Actually you are right, let me do that....

shtukas: @tomrf1 
That one did the trick :)
https://github.com/guardian/price-migration-engine/pull/1099/commits/6e16895fb17d7de51b220b798cb42f3b6842ba32

tomrf1: ah, I was imagining a match on `BillingPeriod` -
```
def billingPeriodToInt(period: BillingPeriod): Int =
  period match {
    case Monthly => 1
    case Quarterly => 3
    case SemiAnnual => 6
    case Annual => 12
  }
```
So that you can call:
`BillingPeriod.billingPeriodToInt(BillingPeriod.fromString(billingPeriod))`

tomrf1 (approved):
👍

PR #1114, opened 2025-02-04
Supporter Plus v1 v2 migration (2025 Edition)

shtukas:
Here we migrate 1000 or so supporter plus subscriptions that are still in version 1 (leftover from the 2023 migration).


shtukas:
Closing this as the migration was done using a Ruby script.

PR #1138, opened 2025-05-30
Decommission GW2024

shtukas:
GW2024 has completed. This decommission the migration code. 

Note that we now no longer have any migration that uses the old / deprecated Zuora Amendment API 🎉


tomrf1 (approved):
🎉

PR #1139, opened 2025-06-09
Introduce migrationExtraAttributes

shtukas:
This introduces a new optional cohort item attribute called `migrationExtraAttributes`. The context for this is explained in the provided documentation.


shtukas:
See missing part in https://github.com/guardian/price-migration-engine/pull/1158

PR #1140, opened 2025-06-09
refactoring: brazeCampaignName to brazeName

shtukas:
This performs a small naming refactoring `brazeCampaignname` to `brazeName` to reflect the fact that nowadays we are mostly using Canvas for engine user communication in Braze. 

Both the code and the documentation have been updated and I have also rebuild the `price-migration-engine-cohort-spec-PROD` table in AWS (there currently is only one migration running)

Screenshot 2025-06-09 at 11 26 19

### DIFF
--- a/docs/migration-implementation-manual.md
+++ b/docs/migration-implementation-manual.md
@@ -19,7 +19,7 @@ As part of setting up a migration you will want to run some of the lambdas on de
 ```
 {
     "cohortName":"GW2024",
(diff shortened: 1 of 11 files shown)


tomrf1 (approved):
👍 if the data is being migrated

shtukas:
@tomrf1 
Yes the data in this case, was just one record of the table that fires up the migrations in the morning. The name is passed to the state machine

PR #1148, opened 2025-06-11
Full implementation of GuardianWeekly2025 amendmentOrderPayload

shtukas:
Here we fully implement GuardianWeekly2025's `amendmentOrderPayload` function, with the updated test. This is the follow of yesterday's ( https://github.com/guardian/price-migration-engine/pull/1145 )


lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
```diff
@@ -165,22 +166,37 @@ object GuardianWeekly2025Migration {
...
+        productRatePlanId = ratePlan.productRatePlanId // We are upgrading on the same rate plan.
+        chargeOverrides = List(
+          ZuoraOrdersApiPrimitives.chargeOverride(
+            ratePlan.ratePlanCharges.headOption.get.productRatePlanChargeId,
+            PriceCap.cappedPrice(oldPrice, estimatedNewPrice, priceCap)
+          )
+        )
+        addProduct = ZuoraOrdersApiPrimitives.addProduct(triggerDateString, productRatePlanId, chargeOverrides)
+        order_subscription = ZuoraOrdersApiPrimitives.subscription(subscriptionNumber, removeProduct, addProduct)
+        o = ZuoraOrdersApiPrimitives.replace_a_product_in_a_subscription(
```
graham228221, line 184:
```suggestion
        orderPayload = ZuoraOrdersApiPrimitives.replace_a_product_in_a_subscription(
```
naming it something like that might be clearer?

graham228221 (approved):
let's go! 📈

lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
shtukas: Thanks for that. I will update :)

shtukas: Done here: https://github.com/guardian/price-migration-engine/pull/1148/commits/78ac905be648d39aacd2d1c8c2ffc18b8ddf43a6 , thanks!

PR #1155, opened 2025-06-19
Implement Newspaper2025P1Migration.getNotificationData

shtukas:
Using the migration extended attributes encoded in the cohort item, we are implementing a feature of Newspaper2025 that will carry the exact label to be used in the communications. This is a value 

```
{ "brandTitle": "the Guardian and the Observer" }
```

that will have to be computed using a script and uploaded to the dynamo table before the notifications start.


lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2025P1Migration.scala
```diff
@@ -75,6 +99,42 @@ object Newspaper2025P1Migration {
...
+          case None =>
+            ZIO.fail(
+              DataExtractionFailure(
+                s"Could not build Newspaper2025P1NotificationData for item ${item.toString}"
+              )
+            )
+          case Some(d) => ZIO.succeed(d)
+        }
+      }
+      case _ => ZIO.succeed(Newspaper2025P1NotificationData(""))
```
tomrf1, line 134: should this ever happen? Is it an error?

tomrf1: I see. Perhaps the intent would be a little clearer if this function returns `ZIO.succeed(None)` in this case - currently you eventually wrap the value in a Some (`newspaper2025_brand_title`) anyway.

tomrf1 (approved):

PR #1159, opened 2025-06-23
activate GuardianWeekly2025 amendments and nexts

shtukas:
This activate the GuardianWeekly2025 amendment and following steps.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
@@ -207,7 +207,7 @@ object GuardianWeekly2025Migration {
 
     val order_opt = {
       for {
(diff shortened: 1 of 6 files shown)


graham228221 (approved):
let's go!

lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
```diff
@@ -207,7 +207,7 @@ object GuardianWeekly2025Migration {
 
     val order_opt = {
       for {
-        ratePlan <- SI2025RateplanFromSubAndInvoices.determineRatePlan(subscription, invoiceList)
+        ratePlan <- SI2025RateplanFromSubAndInvoices.determineRatePlan(zuora_subscription, invoiceList)
```
tomrf1, line 210: `val zuora_subscription = subscription` is a curious line - why not rename the function parameter?

tomrf1 (approved):

PR #1165, opened 2025-06-25
update signature of ZuoraOrdersApiPrimitives.subscription

shtukas:
Here we remove the assumption that when building a `subscription` in the context of the Zuora Order API, that there is only one addition and one removal. This has been the case for far, but for some GuardianWeekly2025 subs, we will be removing 2 rate plans (one of them being a Discount as per previous PR: https://github.com/guardian/price-migration-engine/pull/1163)


lambda/src/test/resources/Migrations/GuardianWeekly2025/6801-Discount/account.json
```diff
@@ -13,8 +13,8 @@
     "purchaseOrderNumber": null,
     "customerServiceRepName": null,
     "ScrubbedOn__c": null,
-    "IdentityId__c": "IDENTITY_ID__c",
-    "sfContactId__c": "sfContactId__c",
+    "IdentityId__c": "100825957",
+    "sfContactId__c": "0030J00001uyVrKQAU",
```
graham228221, line 17: Avoid having real customer data (Identity IDs in particular are considered PII)

```suggestion
    "IdentityId__c": "123456",
    "sfContactId__c": "654321",
```

PR #1175, opened 2025-07-08
HomeDelivery2025: helper functions and tests

shtukas:
Here we :
- refactor the name of the subscription delivery pattern
- Add an migration extra attribute new parameters: `removeDiscount` (will update the cohort items in AWS using a Ruby script)
- Implement the corresponding test functions, including for the price look up

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/HomeDelivery2025Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/HomeDelivery2025Migration.scala
@@ -101,8 +101,8 @@ object HomeDelivery2025Migration {
   def decideShouldRemoveDiscount(item: CohortItem): Boolean = {
     val flag_opt = (for {
       attributes <- item.migrationExtraAttributes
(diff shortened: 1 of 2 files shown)


lambda/src/main/scala/pricemigrationengine/migrations/HomeDelivery2025Migration.scala
```diff
@@ -97,6 +98,16 @@ object HomeDelivery2025Migration {
     }
   }
 
+  def decideShouldRemoveDiscount(item: CohortItem): Boolean = {
+    val flag_opt = (for {
+      attributes <- item.migrationExtraAttributes
+      data: Newspaper2025ExtraAttributes =
+        upickle.default.read[Newspaper2025ExtraAttributes](attributes)
```
graham228221, line 105:
```suggestion
      data: HomeDelivery2025ExtraAttributes =
        upickle.default.read[HomeDeliveryExtraAttributes](attributes)
```

to avoid confusion between HomeDelivery2025 and Newspaper2025 migrations?

PR #1188, opened 2025-07-17
Newspaper2025P1: complete implementation of amendmentOrderPayload (v2)

shtukas:
Here we perform the second part of the implementation of Newspaper2025P1Migration.amendmentOrderPayload (see first part: https://github.com/guardian/price-migration-engine/pull/1182).We add support for subscriptions with a discount which come with an extra attribute instruction of removing it. 

The resulting test is our most complex payload so far and the first time that we have a single payload removing two products (the discount being one of them) and adding one with multiple charges.


lambda/src/main/scala/pricemigrationengine/migrations/Newspaper2025P1Migration.scala
```diff
@@ -266,25 +266,57 @@ object Newspaper2025P1Migration {
...
+          ZuoraOrdersApiPrimitives.replace_a_product_in_a_subscription(
+            orderDate.toString,
+            accountNumber,
+            order_subscription
+          )
+        }
+      } else {
+        for {
+          ratePlan <- SI2025RateplanFromSubAndInvoices.determineRatePlan(zuora_subscription, invoiceList)
+          discount <- GuardianWeekly2025Migration.getDiscount(zuora_subscription)
```
tomrf1, line 294: GuardianWeekly?

shtukas: Ah. Good catch. I corrected it in another change I am about to release, but not here.

PR #1200, opened 2025-08-03
add logging for investigation

shtukas:
Add logging to help with an investigating of GuardianWeekly2025

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
@@ -172,22 +172,35 @@ object GuardianWeekly2025Migration {
   // - amendmentOrderPayload is used in the Amendment handler
   // ------------------------------------------------
 
+  def logValue[T](label: String)(value: T): T = {
+    println(s"$label: $value")
+    value
+  }
+
   def priceData(
       subscription: ZuoraSubscription,
       invoiceList: ZuoraInvoiceList,
       account: ZuoraAccount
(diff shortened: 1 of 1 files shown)


lambda/src/main/scala/pricemigrationengine/migrations/GuardianWeekly2025Migration.scala
```diff
@@ -172,22 +172,36 @@ object GuardianWeekly2025Migration {
...
     val priceDataOpt: Option[PriceData] = for {
-      ratePlan <- SI2025RateplanFromSubAndInvoices.determineRatePlan(subscription, invoiceList)
-      currency <- SI2025Extractions.determineCurrency(ratePlan)
-      oldPrice = SI2025Extractions.determineOldPrice(ratePlan)
-      localisation <- SubscriptionLocalisation.determineSubscriptionLocalisation(subscription, invoiceList, account)
-      billingPeriod <- SI2025Extractions.determineBillingPeriod(ratePlan)
+      _ <- Some(()).map(logValue("initialization"))
+      ratePlan <- SI2025RateplanFromSubAndInvoices
+        .determineRatePlan(subscription, invoiceList)
+        .map(logValue("ratePlan"))
```
tomrf1, line 189: not important, but `tap` might be more idiomatic - https://zio.dev/reference/core/zio/#tapping

shtukas: Oh. Nice one. I saw that being used in the ZIO code, but I didn't know it was basic Scala 🙏

PR #1204, opened 2025-08-10
Decommission migrationCompleteDate

shtukas:
The `CohortSpec` attribute `migrationCompleteDate` is not being used. 

Instead we read the cohort table to know the last migrationDate and check the distinct processing stages of the short table.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
@@ -12,11 +12,10 @@ object MigrationHandler extends ZIOAppDefault with RequestHandler[Unit, Unit] {
 
   private val migrateActiveCohorts =
     (for {
-      today <- Clock.currentDateTime.map(_.toLocalDate)
       cohortSpecs <- CohortSpecTable.fetchAll
       activeSpecs <-
         ZIO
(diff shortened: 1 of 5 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
```diff
@@ -12,11 +12,10 @@ object MigrationHandler extends ZIOAppDefault with RequestHandler[Unit, Unit] {
 
   private val migrateActiveCohorts =
     (for {
-      today <- Clock.currentDateTime.map(_.toLocalDate)
       cohortSpecs <- CohortSpecTable.fetchAll
       activeSpecs <-
         ZIO
-          .filter(cohortSpecs)(cohort => ZIO.succeed(CohortSpec.isActive(cohort)(today)))
+          .filter(cohortSpecs)(cohort => ZIO.succeed(true)) // TODO: simplify this
```
tomrf1, line 18: do we need this line still?

tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/handlers/MigrationHandler.scala
shtukas: @tomrf1 
Technically no, but when I removed it the result didn't compile so I decided to refactor the entire block in another step.

PR #1211, opened 2025-08-15
refactor: Remove tautologous block

kelvin-chappell:


shtukas:
Thank you @kelvin-chappell !

PR #1217, opened 2025-08-20
Async Zuora renewals and product amendments

shtukas:
We migrate the renewal operation as well as the product migration operation, those using the new Orders API payloads, from synchronous to asynchronous processing. This is to avoid the messy timeouts we recently experienced with the print migrations this summer (2025).

nb:  with this change we should now have 100% successful completion rate of the Amendment handler (at least this is what the R&D engine has shown over the past couple of weeks with this method), but we still had timeouts happening during the Estimation step. (Those subs are currently in `EstimationFailed` status). Some future work will try and address that.

ps: as part of this change, we also revert this operation: https://github.com/guardian/price-migration-engine/pull/1209


lambda/src/main/scala/pricemigrationengine/services/Zuora.scala
```diff
@@ -21,15 +21,11 @@ trait Zuora {
...
-      subscription: ZuoraSubscription,
-      payload: Value
-  ): ZIO[Any, ZuoraOrderFailure, Unit]
-
-  def renewSubscription(
+  def applyOrderAsynchronously(
       subscriptionNumber: String,
-      payload: Value
-  ): ZIO[Any, ZuoraRenewalFailure, Unit]
+      payload: Value,
```
kelvin-chappell, line 26: This looks like an established pattern, but I wonder if the payload were a type instead of a Json object, which had a writer so that it could be converted to Json in the service implementation, that might be easier to reason about.  Also then the logging description could be mapped to the payload type and wouldn't need to be passed in here.

lambda/src/main/scala/pricemigrationengine/services/Zuora.scala
```diff
@@ -21,15 +21,11 @@ trait Zuora {
       payload: ZuoraAmendmentOrderPayload
   ): ZIO[Any, ZuoraOrderFailure, Unit]
 
-  def applyAmendmentOrder_json_values(
-      subscription: ZuoraSubscription,
-      payload: Value
-  ): ZIO[Any, ZuoraOrderFailure, Unit]
-
-  def renewSubscription(
+  def applyOrderAsynchronously(
```
kelvin-chappell, line 24: Perhaps some scaladoc to explain what it's for and when to use it.

lambda/src/main/scala/pricemigrationengine/services/ZuoraLive.scala
```diff
@@ -234,60 +234,86 @@ object ZuoraLive {
...
+        private def submitAsynchronousOrderRequest(
+            subscriptionNumber: String,
             payload: Value
-        ): ZIO[Any, ZuoraOrderFailure, Unit] = {
+        ): ZIO[Any, ZuoraAsynchronousOrderRequestFailure, AsyncJobSubmissionTicket] = {
           val payload_stringified = payload.toString()
-          post[ZuoraAmendmentOrderResponse](
-            path = s"orders",
+          post[AsyncJobSubmissionTicket](
+            path = s"async/orders",
```
kelvin-chappell, line 243:
```suggestion
            path = "async/orders",
```

lambda/src/main/scala/pricemigrationengine/services/ZuoraLive.scala
```diff
@@ -234,60 +234,86 @@ object ZuoraLive {
...
             body = payload_stringified
           ).foldZIO(
             failure = e =>
               ZIO.fail(
-                ZuoraOrderFailure(
-                  s"[c37ad58f] subscription number: ${subscription.subscriptionNumber}, payload: ${payload}, reason: ${e.reason}"
+                ZuoraAsynchronousOrderRequestFailure(
+                  s"[c8158a4f] subscription number: ${subscriptionNumber}, payload: ${payload}, reason: ${e.reason}"
```
kelvin-chappell, line 249: What are these codes, like `[c8158a4f]`?  Is that documented somewhere?

kelvin-chappell (approved):
Nice move into async services!  🥳 

I've made a couple of comments, which you can consider or ignore as you please.
But it's a good change.

lambda/src/main/scala/pricemigrationengine/services/Zuora.scala
shtukas: Not exactly scaladoc, but added comments, notably to explain the use of `operationDescriptionForLogging`.

PR #1221, opened 2025-08-23
Improve Zuora data retrieval

shtukas:
In this PR we move from this amendment logic (a simplified view of the operations of the Amendment handler)

```
(...)

_ <- Zuora.renewSubscription(subscription.number)

invoicePreview <- Zuora.fetchInvoicePreview(subscription.accountNumber)

payload <- ZIO.fromEither(buildOrderPayload(subscription, invoicePreview))

_ <- Zuora.submitPayloadToOrdersAPI(payload)

(...)

```

to

```
(...)

_ <- Zuora.renewSubscription(subscription.number)

payload <- (for {
    invoicePreview <- Zuora.fetchInvoicePreview(subscription.accountNumber)
    payload <- ZIO.fromEither(buildOrderPayload(subscription, invoicePreview))
} yield payload).retry(exponential(10.second) && recurs(6))

_ <- Zuora.submitPayloadToOrdersAPI(payload)

(...)
```

This is to address a small issue with Zuora by which, in rare occasions, the invoice preview retrieved just after the renew operation fails to be the correct one the build the product amendment payload. This is consistent with previous observations that the primary and secondary side effects of some operations are not flowing throw instantaneously within Zuora.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -327,7 +327,7 @@ object AmendmentHandler extends CohortHandler {
               )
             )
         }
-      } yield order).retry(exponential(1.second) && recurs(5))
+      } yield order).retry(exponential(10.second) && recurs(6))
 
       _ <- Logging.info(
         s"[6e6da544] Amending subscription ${subscriptionBeforeUpdate.subscriptionNumber} with order ${order}"
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -248,6 +248,8 @@ object AmendmentHandler extends CohortHandler {
         subscriptionBeforeUpdate.subscriptionNumber
       )
 
+      _ <- renewSubscription(subscriptionBeforeUpdate, subscriptionBeforeUpdate.termEndDate, account)
+
       order <- (for {
         _ <- Logging.info(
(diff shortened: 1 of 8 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -248,74 +250,85 @@ object AmendmentHandler extends CohortHandler {
...
+                zuora_subscription = subscriptionBeforeUpdate,
+                oldPrice = oldPrice,
+                estimatedNewPrice = estimatedNewPrice,
+                priceCap = Newspaper2025P3Migration.priceCap,
+                invoiceList = invoicePreviewBeforeUpdate
+              )
             )
-          )
-      }
+        }
```
kelvin-chappell, line 329: I think I would extract this section as it looks like pure code that could be entirely wrapped in a `ZIO.fromEither`.  Also having an `order` value shadowing an outer `order` value might be confusing.

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -248,74 +250,85 @@ object AmendmentHandler extends CohortHandler {
...
+      } yield order).retry(exponential(10.second) && recurs(6))
```
kelvin-chappell, line 330: 💯

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
shtukas: Good idea! Will do

PR #1222, opened 2025-08-24
Better time management for the Amendment handler

shtukas:
This is a follow up PR for https://github.com/guardian/price-migration-engine/pull/1221 (retry of the invoice preview retrieval).

At 2025-08-24T03:20:02.127000+00:00 in the logs, there was an item that took 13 minutes to complete and although it then ended up being migrated just fine in Zuora, this caused the lambda to exceed 15 minutes runtime and be terminated by the AWS environment. 

In this change we make the Amendment handler better at time management.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -24,9 +24,23 @@ object AmendmentHandler extends CohortHandler {
 
   private def main(cohortSpec: CohortSpec): ZIO[Logging with CohortTable with Zuora, Failure, HandlerOutput] = {
     for {
+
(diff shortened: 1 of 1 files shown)


shtukas:
ps: A cohort item that goes through `ZuoraAmendmentPayloadBuildingFailure` will be counted as having been processed. This might create the following situation:

Let's say that 14 items were scheduled to be amended today. Note that currently the batch size is 15. Assume that 13 items are amended fine, but the last one doesn't get the right invoice preview on time before the retry schedule exits. Then the lambda will return `Output: HandlerOutput(true)`

This is not a problem because then the item will be processed the day after. 

(Moreover, Although the AWS step function is scheduled to run once a day, Pascal currently manually runs the state machine several times a day, so even the late items get to be amended on the first day.)

PR #1227, opened 2025-08-27
Extend CohortItems - Part 1: new attributes

shtukas:
Earlier this year, we introduced to encode migration specific attributes, to support migration specific features, as a serialized json object stored in the field of a Dynamo table. This has worked very well, and as a prelude to ProductMigration2025N4, we refactor that scheme by introducing new attributes that we are going to use to unfold the json objects. This will make cohort items much easier to use as we extend their capabilities.

This is the first of three changes. In change 2 we are going to refactor (unfold) the current print migrations `migrationExtraAttributes` into the new flat attributes. In change 3 we are going to decommission `migrationExtraAttributes`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ It currently consists in
     - [Downloading fixtures](docs/downloading-fixtures.md)
     - [Notes on the Zuora Order API](docs/zuora-order-api.md)
     - [Communication with braze](docs/communication-with-braze.md)
-    - [Extending cohort items](docs/extending-CohortItem.md)
+    - [Cohort Items](docs/cohort-item.md)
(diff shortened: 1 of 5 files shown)


shtukas:
Talked with @tomrf1  and we decided to use the method and conventions we are using for email messages: https://github.com/guardian/price-migration-engine/blob/c11bb73debbbeab3387ea28dd88144c1bc084e76/lambda/src/main/scala/pricemigrationengine/model/membershipworkflow/EmailMessage.scala#L23

tiny bit more management of the Scala type itself, but safer when reading and maintaining the code. And in particular there will be mo need to maintain a document mapping for the meaning each migration gives to each key.

PR #1233, opened 2025-09-13
Upgrade signature of ZuoraOrdersApiPrimitives.ratePlanChargesToChargeOverrides

shtukas:
Context: Newspaper2025P1, HomeDelivery2025 and Newspaper2025P3

As part of the operations of the Amendment handler, the function `ZuoraOrdersApiPrimitives.ratePlanChargesToChargeOverrides` generates charge overrides that correspond to the total estimated new price. 

Each individual price charge is a multiplicative increase of the old price that is then rounded to precision 0.01. Sadly this means that floating point errors can then accumulate to the total price of the amended subscription, which then can be a few pennies less than the originally estimated price rise. This was not detected by the post amendment price check which only checks that the new price is not higher than the estimated new price. [1]

Here we update `ratePlanChargesToChargeOverrides` as follow:
- The target price, the estimated new price is passed to the function
- We ensure (possibly performing price adjustment) that the total prices of the charge overrides is exactly the value that was passed. When an adjustment is needed it is carried by the first leg, that was an arbitrary choice. 

[1] A reenforced price check was released as a companion PR here: https://github.com/guardian/price-migration-engine/pull/1234

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/HomeDelivery2025Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/HomeDelivery2025Migration.scala
@@ -236,6 +236,7 @@ object HomeDelivery2025Migration {
           val chargeOverrides: List[Value] = ZuoraOrdersApiPrimitives.ratePlanChargesToChargeOverrides(
             ratePlan.ratePlanCharges,
             priceRatio,
+            estimatedNewPrice,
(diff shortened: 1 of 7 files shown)


shtukas:
Something I should probably do in the future is treat all the prices in the engine as multiples of 0.01, in which case they can all be `Long` instead of the `BigDecimal` we are using for essentially decimal numbers.

PR #1234, opened 2025-09-13
strengthen the post amendment price check

shtukas:
Traditionally the engine has been tolerating of final prices that are a penny (or cent) lower than the estimated new price. Here we make that check stronger and require equality. This might cause the engine to fail in situation Zuora would perform float operations different of that of the engine, but we will investigate every such case and update the migration code accordingly. 

Note that this is a companion PR of: https://github.com/guardian/price-migration-engine/pull/1233

Update:

Discounts on a subscription do not play well with the price post amendment, and the equality check will fail when that happens. The next time it happens I will check the sub for active discounts and will then update the code to behave as follows:
- If the post amendment sub doesn't have discounts, then perform the price equality test.
- If the post amendment sub has discounts then perform the inequality test.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -417,7 +417,7 @@ object AmendmentHandler extends CohortHandler {
         )
 
       _ <-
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):

shtukas:
Part 2 (after the alarm that we were waiting for): https://github.com/guardian/price-migration-engine/pull/1237

PR #1238, opened 2025-09-22
Ph 20250922 1759 salesforce opt out

shtukas:
Migration ProductMigration2025N4 is going to use a new attribute of `SalesforcePriceRise`, which itself needs to be retrieved from Salesforce. @graham228221 did the work of upgrading Salesforce, and here we

1.  Upgrade `SalesforcePriceRise` with `Customer_Opt_Out__c`

2. Update `SalesforceClient` with `getPriceRise(priceRiseId: String): IO[SalesforceClientFailure, SalesforcePriceRise]`

3. Temporarily update SalesforceAmendmentUpdateHandler.scala, to call this new end point.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceAmendmentUpdateHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceAmendmentUpdateHandler.scala
@@ -36,6 +36,7 @@ object SalesforceAmendmentUpdateHandler extends CohortHandler {
         ZIO
           .fromOption(item.salesforcePriceRiseId)
           .orElseFail(SalesforcePriceRiseWriteFailure("salesforcePriceRiseId is required to update Salesforce"))
+      _ <- SalesforceClient.getPriceRise(salesforcePriceRiseId)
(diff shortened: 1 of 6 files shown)


shtukas:
@graham228221 

I tested several time with a patches like this

```
curl -X PATCH \
    "https://gnmtouchpoint.my.salesforce.com/services/data/v64.0/sobjects/Price_Rise__c/[removed]/" \
    -H "Authorization: Bearer [removed]" \
    -H "Content-Type: application/json" \
    -d '{
    "Cancellation_Reason__c": "[random]",
}'
```
and this

```
curl -X PATCH \
    "https://gnmtouchpoint.my.salesforce.com/services/data/v64.0/sobjects/Price_Rise__c/[removed]/" \
    -H "Authorization: Bearer [removed]" \
    -H "Content-Type: application/json" \
    -d '{
    "Cancellation_Reason__c": "[random]",
    "Customer_Opt_Out__c": true
}'
```

And ran a GET afterwards. The conclusion is that Salesforce only update the keys that were provided in the PATCH payload. In particular any existing value of `Customer_Opt_Out__c` is not updated unless it was specifically set in the payload.

PR #1241, opened 2025-09-25
Deserialise SalesforceClientLive

shtukas:
The change we released here ( https://github.com/guardian/price-migration-engine/pull/1238 ) [ SalesforceClient.getPriceRise ], is causing 

```
failed to deserialise: upickle.core.AbortException: expected string got float64 at index 576
```

The problem is caused by the `SalesforcePriceRise` case class defining the attribute

```
Guardian_Weekly_New_Price__c: Option[BigDecimal]
```

But the JSON object returned from Salesforce presenting it as 

```
"Guardian_Weekly_New_Price__c": 335.88
```

Turns out that in uPickle, the default deserialization rules for BigDecimal expect a JSON string, not a raw JSON number.

Here we add support for deserializing a BigDecimal, from a JSON float.


graham228221 (approved):

shtukas:
Successful operation:

2025-09-25T13:04:52.988000+00:00 2025/09/25/[$LATEST]

{"Build":"2518","CohortName":"Newspaper2025P3","INFO":"[774f676b] Successfully retrieved Salesforce price rise object, priceRiseId: [removed], priceRise: SalesforcePriceRise(Some([removed]),Some([removed]),Some(311.88),Some(335.88),Some(2025-10-26),Some([removed]),Some(2025-09-21),Some([removed]),Some(Newspaper2025P3),Some(AmendmentComplete),None,Some(false))"}

PR #1257, opened 2025-10-02
refactorings

shtukas:
Some refactoring prelude for https://github.com/guardian/price-migration-engine/pull/1258

Note an update in the shape of `SuccessfulAmendmentResult`. It was unfitting that it would carry the `oldPrice` and the `estimatedNewPrice`. It's now more fit for purpose.


lambda/src/main/scala/pricemigrationengine/model/CohortItem.scala
```diff
@@ -8,24 +8,53 @@ import java.time.{Instant, LocalDate}
...
+    // Price derived from the Estimation step, without capping
     estimatedNewPrice: Option[BigDecimal] = None,
+
+    // Price (with possible capping) used in the communication to the user and sent to Salesforce
+    commsPrice: Option[BigDecimal] = None,
+
+    // Price read from the post amendment subscription
+    newPrice: Option[BigDecimal] = None,
+
+    //
```
tomrf1, line 25: no comment?

tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/model/CohortItem.scala
shtukas: Yeah. It's to force a separation (stylistic presentation of the code). Otherwise the linter (scalafmt) remove the space :)

PR #1451, opened 2026-06-30
Introduce SupporterPlus2026

shtukas:
Introducing `SupporterPlus2026` as new MigrationType. This is the first step towards the supporter plus migration starting this summer.


andrade0 (approved):
LGTM

PR #1452, opened 2026-06-30
renaming and linting

shtukas:
Here we 

1. Stop using the name `amendmentMigrationDispatch` which I had introduced when we were still using two (the old and the new orders api) amendment API. 

2. Improve formatting by renaming `doAmendment_ordersApi_json_values`

3. Add and improve a few comments.


andrade0 (approved):
LGTM

PR #1454, opened 2026-07-02
date calculations for SupporterPlus2026

shtukas:
Here we implement the non standard date calculations for SupporterPlus2026. Notably the novel idea of migrating the monthlies over 6 weeks.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/model/AmendmentEffectiveDateCalculator.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/AmendmentEffectiveDateCalculator.scala
@@ -1,7 +1,12 @@
 package pricemigrationengine.model
 
 import pricemigrationengine.handlers.NotificationHandler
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/model/AmendmentEffectiveDateCalculator.scala
```diff
@@ -159,6 +164,20 @@ object AmendmentEffectiveDateCalculator {
     // Decides an integer in the interval [0, spreadPeriod-1]
     // The default spread period is 1
 
-    lowerBound4.plusMonths(randomDelayInMonths)
+    var lowerBound5 = lowerBound4.plusMonths(randomDelayInMonths)
```
enesihsa, line 167: Any reason this is var and not val?

shtukas: Will correct, thanks!

PR #1465, opened 2026-07-05
notification: check product name

shtukas:
There currently is a problem with the Notification handler: if the subscription's active rate plan has moved from one product to another, then the notification would happen erroneously. This can happen with Supporter Plus transmuting to Digital Packs

So we introduce a few functions  

1. `expectedProductNameOpt` in the NotificationHandlerHelper indicate which product name we are expecting a subscription active rate plan to have at the time of notification. Note that it is likely that we are going to slightly update the signature of this function in the future to return a list of names instead of just one. 

2. We add a function in SubscriptionIntrospection2025 to simplify the lookup.

3. Modify the Notification handler `main` function to perform the check and if the subscription fails it, it's moved to a new terminal processing stage called `ExcludedFromMigration`.

Note that there will be another PR where we do the same for the Amendment handler.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
@@ -60,9 +60,31 @@ object NotificationHandler extends CohortHandler {
               .fetch(SalesforcePriceRiseCreationComplete, Some(today.plusDays(maxLeadTime(cohortSpec))))
               .filter(item => item.subscriptionName == subscriptionNumber)
         }
-      )
-        .mapZIO(item => sendNotification(cohortSpec)(item, today))
-        .runCount
+      ).mapZIO { item =>
+        for {
+          subscription <- Zuora.fetchSubscription(item.subscriptionName)
+          hasCorrectProduct = SI2025RateplanFromSub
+            .uniquelyDeterminedActiveNonDiscountNonExpiredRatePlanHasGivenProductNameDefaultToTrue(
+              subscription,
+              today,
+              NotificationHandlerHelper.expectedProductNameOpt(cohortSpec)
+            )
+          result <-
+            if (hasCorrectProduct) {
+              sendNotification(cohortSpec)(item, today)
+            } else {
+              CohortTable
+                .update(
+                  CohortItem(
+                    item.subscriptionName,
+                    processingStage = ExcludedFromMigration,
+                    cancellationReason = Some("excluded from migration due to unexpected product name")
+                  )
+                )
+                .as(())
+            }
+        } yield result
+      }.runCount
     } yield HandlerOutput(isComplete = count < batchSize)
   }
 
--- a/lambda/src/main/scala/pricemigrationengine/model/SubscriptionIntrospection2025.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/SubscriptionIntrospection2025.scala
@@ -102,6 +102,32 @@ object SI2025RateplanFromSub {
         )
     }
   }
+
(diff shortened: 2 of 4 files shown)


lambda/src/main/scala/pricemigrationengine/model/SubscriptionIntrospection2025.scala
```diff
@@ -102,6 +102,32 @@ object SI2025RateplanFromSub {
         )
     }
   }
+
+  def uniquelyDeterminedActiveNonDiscountNonExpiredRatePlanHasGivenProductNameDefaultToTrue(
```
tomrf1, line 106: Just naming, but I'd have gone for something like `hasExpectedProductName` and used a sentence in a comment to cover the rest!

tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/model/SubscriptionIntrospection2025.scala
shtukas: Ah yeah. I realise the name is long, but it's a little bit of a convention inside `SubscriptionIntrospection2025` (mostly to avoid surprises). 

`hasExpectedProductName` is too short, because the name needs to highlight that it needs to be true if `productNameOpt` is `None` (the `true` value in that case is counter intuitive but needed for backward compatibility), it also needs to highlight that it will return `false` if the product name is present but on an expired rate plan, and it also needs to highlight that the return value is going to be false if the product name is present but there are other active rate plans 😅

shtukas: What I _could_ do though, is split the function in two parts, the resolution of the rate plan, and then the check of the rate plan name (with all the right edge cases conventions) in different functions. Will refactor soon 🙂

lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala
```diff
@@ -60,9 +60,31 @@ object NotificationHandler extends CohortHandler {
...
+          result <-
+            if (hasCorrectProduct) {
+              sendNotification(cohortSpec)(item, today)
+            } else {
+              CohortTable
+                .update(
+                  CohortItem(
+                    item.subscriptionName,
+                    processingStage = ExcludedFromMigration,
+                    cancellationReason = Some("excluded from migration due to unexpected product name")
```
johnduffell, line 81: consider to include the actual and expected product names in this message for better debugging?

lambda/src/main/scala/pricemigrationengine/model/SubscriptionIntrospection2025.scala
```diff
@@ -102,6 +102,32 @@ object SI2025RateplanFromSub {
...
+      subscription: ZuoraSubscription,
+      today: LocalDate,
+      productNameOpt: Option[String]
+  ): Boolean = {
+    // This function essentially returns `true` if the uniquely contextual
+    // rate plan product name is what we expect. This was introduced to
+    // ensure that at Notification time or Amendment time the subscription
+    // has not moved to a different product. This can happen to, for instance,
+    // to Supporter Plus subs that can be transmuted to Digital Packs
```
johnduffell, line 115: you can just say "switched" as that's the teminology we use in the product-switch-api

johnduffell (commented):
thanks for fixing 🥇

lambda/src/main/scala/pricemigrationengine/model/SubscriptionIntrospection2025.scala
shtukas: Yeah I guess `switched` is more canonical than `transmuted`.

PR #1470, opened 2026-07-08
SP2026: SP2026EmailExtraAttributes

shtukas:
Agreed with Marketing to add the following extra email parameters for the Braze template. 

```
sp2026_contribution_amount
sp2026_current_combined_amount
sp2026_new_combined_amount
```

We haven't yet build the email payload, but as a stepping stone, we introduce `SP2026EmailExtraAttributes` to carry the information and helper functions to generate the data.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
@@ -9,6 +9,12 @@ import zio.ZIO
 
 import scala.util.Random
 
+case class SP2026EmailExtraAttributes(
+    contributionAmount: BigDecimal,
+    currentCombinedAmount: BigDecimal,
+    newCombinedAmount: BigDecimal
+)
+
 object SupporterPlus2026Migration {
 
   // ------------------------------------------------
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
```diff
@@ -77,6 +83,33 @@ object SupporterPlus2026Migration {
...
+    } yield price
+  }
+
+  def extractEmailExtraAttributes(
+      subscription: ZuoraSubscription,
+      cohortItem: CohortItem
+  ): Option[SP2026EmailExtraAttributes] = {
+    for {
+      contributionAmount <- subscriptionToContributionAmount(subscription)
+      currentBaseAmount <- cohortItem.oldPrice
```
johnduffell, line 102: just adding a link to where oldPrice is calculated - it specifically excludes the contribuition amount hence why we filter it separately
ref: https://github.com/guardian/price-migration-engine/blob/0452675fe066a8410b26816fdd4d8e460b8a2dae/lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala#L97

johnduffell (approved):
looks good thanks for keeping it small and easy to review!

PR #1473, opened 2026-07-09
Configure notification step for SP2026

shtukas:
The email payload for SupporterPlus2026, uses 3 non standard attributes

```
sp2026_contribution_amount
sp2026_current_combined_amount
sp2026_new_combined_amount
```

Based on the work we did to introduce the class `SP2026EmailExtraAttributes` ( yesterday: https://github.com/guardian/price-migration-engine/pull/1470 ), we now extends the email payload. 

Note that

1. I modified `SP2026EmailExtraAttributes` to carry strings, it's then much more natural and easier to use the data for the payload, and also aligned with previous convention of doing so.

2. `brazeName` also now takes a Zuora subscription, this is somehow a departure from what I would have done last couple of year, pre compute the values and add them to the cohort item itself as extra attributes. 

3. The comment "For Tom reading this..." is now a tradition, it's to prevent another discussion about why I can't use optional values, for the fields of `SP2026EmailExtraAttributes` and returning empty strings) for when the function `extractEmailExtraAttributes` is called from other migrations). ( prior art: https://github.com/guardian/price-migration-engine/blob/8ee3463489064e5ce3bcee348488ecd2b7d53677/lambda/src/main/scala/pricemigrationengine/migrations/ProductMigration2025N4Migration.scala#L52 )

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
+++ b/lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
@@ -10,9 +10,9 @@ import zio.ZIO
 import scala.util.Random
 
 case class SP2026EmailExtraAttributes(
-    contributionAmount: BigDecimal,
-    currentCombinedAmount: BigDecimal,
-    newCombinedAmount: BigDecimal
+    contributionAmount: String,
+    currentCombinedAmount: String,
+    newCombinedAmount: String
 )
 
 object SupporterPlus2026Migration {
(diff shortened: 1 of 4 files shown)


lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
```diff
@@ -94,20 +94,75 @@ object SupporterPlus2026Migration {
...
+        } yield SP2026EmailExtraAttributes(
+          contributionAmount.toString(),
+          currentCombinedAmount.toString(),
+          futureCombinedAmount.toString()
+        )
+      }
+      case _ =>
+        Some(
+          // Date: 9th July 2026
+          // For Tom reading this... Same as usual, I can't return a None
```
tomrf1, line 119: 😆

lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
```diff
@@ -10,9 +10,9 @@ import zio.ZIO
 import scala.util.Random
 
 case class SP2026EmailExtraAttributes(
-    contributionAmount: BigDecimal,
-    currentCombinedAmount: BigDecimal,
-    newCombinedAmount: BigDecimal
+    contributionAmount: String,
```
tomrf1, line 13: these fields are for displaying in emails only, so String should be fine 👍

tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/migrations/SupporterPlus2026Migration.scala
shtukas: Correct. When I introduced the `SP2026EmailExtraAttributes` type yesterday I gave it the natural type for the information it carries (that was a change without context), but since it's now going to be used for email payload in field that expect string, I back ported the change of type. If I had not changed the type, I would have needed to return the "neutral" version of a `BigDecimal` , meaning  `BigDecimal(0.0)`, which is not as cool as empty string 🙂

PR #1474, opened 2026-07-09
decommission brazeName from CohortSpec

shtukas:
The fact that cohort specs carry a field called braze name is a vestige from an old era before the modern style of defining migrations (nowadays we put the parameters in the migration's own module -- for instance price grids, notification windows etc). Moreover we no longer use just one canvas template in Braze that led to the introduction of brazeName ( https://github.com/guardian/price-migration-engine/blob/8ee3463489064e5ce3bcee348488ecd2b7d53677/lambda/src/main/scala/pricemigrationengine/handlers/NotificationHandler.scala#L565 ) in the Notification handler. 

Here we decommission brazeName from the cohort specs and to do so in a backward compatible way we record the names for the three oldest currently running migrations `GuardianWeekly2025`, `Newspaper2025P1`  and `Newspaper2025P3`

### DIFF
--- a/docs/amendment-effective-date-computation.md
+++ b/docs/amendment-effective-date-computation.md
@@ -15,7 +15,6 @@ First we need a cohort spec. Let's assume that the cohort spec is
 ```
 {
     "cohortName":"GW2024",
-    "brazeName":"SV_GW_PriceRise2024",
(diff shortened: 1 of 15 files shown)


andrade0 (approved):
LGTM

PR #1475, opened 2026-07-09
coding directive #3

shtukas:
Introducing Coding Directive #3 to explain the pattern used in the Notification handler, and notably the way we return Zero values for email payload parameters, for instance as done here: https://github.com/guardian/price-migration-engine/pull/1473


andrade0 (approved):
LGTM

PR #1477, opened 2026-07-10
date calculations code refactoring

shtukas:
Notably, move the SupporterPlus2026 date calculations to the migration's own module


andrade0 (approved):
LGTM

PR #1478, opened 2026-07-10
product switch 1y policy

shtukas:
Here we implement the requirement to apply a one year delay policy after the "latest product switch", where "latest product switch" is defined that the earliest date the sub has been running this particular product name, rate plan name pair.

It's a stronger condition than global 1y policy.


andrade0 (approved):
LGTM

PR #1485, opened 2026-07-15
SP2026: clarify base amount

shtukas:
Slight clarification of the way we compute the SP2026 email attributes. We should compute the `currentBaseAmount` from the subscription itself rather than the cohort item to prevent any problem, notably when there is a non zero contribution


andrade0 (approved):
LGTM

PR #1486, opened 2026-07-15
SP2026: remove street requirement

shtukas:
This is a similar requirements relaxation than the mailing address we did a bit earlier today ( https://github.com/guardian/price-migration-engine/pull/1484 )

PR #1487, opened 2026-07-15
SP2026: remove country requirement

shtukas:
This is a similar requirements relaxation we did a bit earlier today 

mailing address: https://github.com/guardian/price-migration-engine/pull/1484 
street: https://github.com/guardian/price-migration-engine/pull/1486

Here we remove requirement for country.


andrade0 (approved):
LGTM

PR #1489, opened 2026-07-17
prevent post amendment price check for SP2026

shtukas:
The post amendment price check is being disabled for SupporterPlus2026 because that check compares the comms price (which in this case is the base price) and the post amendment price, but that doesn't work in the case of supporter plus subs with non zero extra contribution.

(For subscriptions with extra contribution the final price is the comms price plus the extra contribution)


andrade0 (approved):
LGTM

PR #1490, opened 2026-07-19
remove unused code (zuora catalogue)

shtukas:
The Zuora catalogue has not been used as part of migrations for years. Here we stop downloading it from Zuora.


andrade0 (approved):
LGTM

PR #1491, opened 2026-07-19
update state machine to rerun UpdatingSalesforceWithNotifications when not done

shtukas:
Before SupporterPlus2026 we never had a migration that sends more than 1000 user notification per day and we have a hard coded limit here: https://github.com/guardian/price-migration-engine/blob/2a7691dd6dd9f3831b684a43d2dc9b0785d0993b/lambda/src/main/scala/pricemigrationengine/handlers/SalesforceNotificationDateUpdateHandler.scala#L12 for one single run of the SalesforceNotificationDateUpdateHandler , and I never noticed that the state machine doesn't restart that lambda if it  has not done all the records for the day.

.

BEFORE (Current PROD):

.

Screenshot 2026-07-19 at 23 36 56

.

AFTER (CODE with the change):

.

Screenshot 2026-07-19 at 23 43 57

### DIFF
--- a/stateMachine/cfn/cfn.yaml
+++ b/stateMachine/cfn/cfn.yaml
@@ -322,7 +322,7 @@ Resources:
                 "Resource": "${UpdatingSalesforceWithNotificationsLambdaArn}",
                 "InputPath": "$.cohortSpec",
                 "ResultPath": "$.result",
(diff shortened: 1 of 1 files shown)


andrade0 (approved):
LGTM

PR #1500, opened 2026-07-30
SupporterPlus2026 Dispatch

shtukas:
### Problem

SupporterPlus2026 is a record breaking price migration with a cohort table containing more than 350,000 items, which is more than 3 times the previous largest migration. One interesting side effect of this large number is that at the moment (and this will continue for an entire month) the daily load of Braze user notification and Zuora amendments takes more than 24 hours. This is due to the fact that we process cohort items one by one. 

### Solutions

There are 3 ways to deal with this problem. 

1. Modify the code to perform some operations in parallel (for instance the amendments). This is not as safe as what it may first looks like because the engine-Zuoro amendment exchange is not idempotent and the ZIO handling of failures would make it difficult to deal with.

2. Introduce new `MigrationType`s and split the existing cohort table, thereby splitting this one migrations into several. I am not doing that because I don't really want to go into the trouble to split the main table.

3. Introduce new `MigrationType`s and use a dispatch function to target the same table from different names. This has never been done before, but is a very clean solution. 

### Adopted solution

Here we are doing solution 3. I introduced 4 new migration names, namely: `SupporterPlus2026N2`, `SupporterPlus2026N3`, `SupporterPlus2026N4` and `SupporterPlus2026N5`, to bring the number of names targeting the SupporterPlus2026 dynamo table to 5. 

Then we introduce the Dispatch helper, and the `evaluate` function is the core of it. It takes a string (essentially a subscription number) and an integer between 1 and 5 (we have 5 names for SupporterPlus2026) and return a boolean. The requirements are

1. For any string, there must be an integer n such that `evaluate(string, n)` is true
2. For any string, there is a unique integer n such `evaluate(string, n)` is true

This implies that the space of strings is partitioned into 5 subsets

Note that we needed to hack the function that turn a cohort name into a table name, so that all the `SupporterPlus2026*` go to `PriceMigration-PROD-SupporterPlus2026`

The rest is then to plug the dispatch to the handlers, thereby affecting all the migrations, although the effect will only matter for SupporterPlus2026. In any case, we will now have 5 different state machines targeting the same cohort table without any conflicts.

### Deployment

Once the PR is merged, I am going to add the missing entries to the `price-migration-engine-cohort-spec-PROD` Dynamo table, to active the new state machines. 

### Future development

This is a temporary solution to help SupporterPlus2026. It is likely that I am going to decommission Dispatch in a month, but I have documented the solution for if we have another big migration at some point in the future.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/model/Dispatch.scala
+++ b/lambda/src/main/scala/pricemigrationengine/model/Dispatch.scala
@@ -1,0 +1,43 @@
+package pricemigrationengine.model
+
+import scala.util.hashing.MurmurHash3
+
+object Dispatch {
+
+  /*
+    The Displach functionality was introduced to help with SupporterPlus2026
+    https://github.com/guardian/price-migration-engine/pull/1500
+
+    ... and the `evaluate` function is the core of it. It takes a string (essentially a subscription number)
+    and an integer between 1 and 5 (we have 5 names for SupporterPlus2026) and return a boolean.
+
+    The requirements are
+    1. For any string, there must be an integer n such that evaluate(string, n) is true
+    2. For any string, there is a unique integer n such evaluate(string, n) is true
+
+    This implies that the space of strings is partitioned into 5 subsets
+   */
+  def evaluate(s: String, n: Int): Boolean = {
+    require(n >= 1 && n <= 5, "n must be between 1 and 5")
+    val bucket = Math.abs(MurmurHash3.stringHash(s) % 5)
+    bucket == n - 1
+  }
+
+  /*
+    This function returns whether the cohort item "belongs" to the migration type
+    It defaults to true for non SupporterPlus2026 migrations, but induces a partition
+    in 5 subsets for SupporterPlus2026
+   */
+  def belongs(cohortSpec: CohortSpec, cohortItem: CohortItem): Boolean = {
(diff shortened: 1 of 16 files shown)


kelvin-chappell (approved):
🚀 
I wonder if there's a way to generalise this behaviour so that you look at the size of a cohort and then kick off the right number of state machines each with some subset of the cohort routed to it.

tomrf1 (approved):

lambda/src/main/scala/pricemigrationengine/model/Dispatch.scala
```diff
@@ -0,0 +1,43 @@
+package pricemigrationengine.model
+
+import scala.util.hashing.MurmurHash3
+
+object Dispatch {
+
+  /*
+    The Displach functionality was introduced to help with SupporterPlus2026
```
johnduffell, line 8:
```suggestion
    The Dispatch functionality was introduced to help with SupporterPlus2026
```

shtukas: Thanks, I corrected it here: https://github.com/guardian/price-migration-engine/pull/1507/changes/a8e745e0eb3b966f656a2e8e2d0512265e20101a

PR #1507, opened 2026-08-04
amendment handler refactoring

shtukas:
A few days ago ( https://github.com/guardian/price-migration-engine/pull/1504 ) we updated the Notification handler to adopt a simpler and more logical preambule:  [data collection] -> [analysis]  -> [evaluation of the analysis result].

Here we do the same for the amendment handler.


lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -21,29 +22,19 @@ import ujson._
...
-      // which should be plenty of time (observations show that it runs in 4 minutes in average)
-      // but the refactoring we made here: https://github.com/guardian/price-migration-engine/pull/1221
-      // (retry of the invoice preview retrieval) means that we have much less control
-      // over the total time taken by the run.
-
-      // Specifically in one instance an item took 13 minutes to complete, which caused issues
-      // for the lambda itself. To address this problem we monitor the time since the start
-      // of the run, and exit after 10 minutes.
-
       startingTime <- Clock.nanoTime
```
johnduffell, line 33: fyi the Context object of the lambda has a time remaining method that you can use, it always matches with the configured value in the AWS 
https://docs.aws.amazon.com/lambda/latest/dg/java-context.html#:~:text=Context%20methods-,getRemainingTimeInMillis,-()%20%E2%80%93%20Returns%20the

shtukas: Ah, Nice one. I didn't know that.

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -63,20 +54,75 @@ object AmendmentHandler extends CohortHandler {
...
+
+  def evaluateAnalyseResult(
+      cohortSpec: CohortSpec,
+      item: CohortItem,
+      zuoraSubscription: ZuoraSubscription,
+      analyseResult: SubscriptionAmendmentAnalyseResult
+  ): ZIO[CohortTable with SalesforceClient with Logging with Zuora, Failure, Unit] = {
+    analyseResult match {
+      case SAARReadyToAmend =>
+        performAmendmentAttempt(
```
johnduffell, line 101: minor one but since we always call CohortTable.update, it would be good to extract that, and call it outside of this match statement.  At first glance it looked like we only update the cohort table if it's not ready to amend, but digging down, it's a necessary step in all cases (other than an unexpected error)

lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -259,3 +259,24 @@ object AmendmentHandlerHelper {
...
+
+object SAARReadyToAmend extends SubscriptionAmendmentAnalyseResult
+object SAARCancelledInZuora extends SubscriptionAmendmentAnalyseResult
+object SAARExcludeFromMigration extends SubscriptionAmendmentAnalyseResult
+
+object SubscriptionAmendmentAnalyseResult {
+  def analyseSubscriptionForAmendment(
+      item: CohortItem,
+      subscription: ZuoraSubscription
+  ): Option[SubscriptionAmendmentAnalyseResult] = {
```
johnduffell, line 275: why does it return an option when it's always Some?

lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -259,3 +259,24 @@ object AmendmentHandlerHelper {
...
+
+// "SAAR" means "Subscription Amendment Analyse Result"
+
+object SAARReadyToAmend extends SubscriptionAmendmentAnalyseResult
+object SAARCancelledInZuora extends SubscriptionAmendmentAnalyseResult
+object SAARExcludeFromMigration extends SubscriptionAmendmentAnalyseResult
+
+object SubscriptionAmendmentAnalyseResult {
+  def analyseSubscriptionForAmendment(
+      item: CohortItem,
```
johnduffell, line 273: this doesn't seem to need `item`

lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -259,3 +259,24 @@ object AmendmentHandlerHelper {
...
+sealed trait SubscriptionAmendmentAnalyseResult
+
+// "SAAR" means "Subscription Amendment Analyse Result"
+
+object SAARReadyToAmend extends SubscriptionAmendmentAnalyseResult
+object SAARCancelledInZuora extends SubscriptionAmendmentAnalyseResult
+object SAARExcludeFromMigration extends SubscriptionAmendmentAnalyseResult
+
+object SubscriptionAmendmentAnalyseResult {
+  def analyseSubscriptionForAmendment(
```
johnduffell, line 272: can we call this something like SubscriptionNextStage.fromSubscription as it's shorter and also slightly more concise?  I'm not keen on the `SAAR` shortening, but I agree it would be too verbose otherwise.
A trick I would recommend is to move the three case objects above inside `object SubscriptionAmendmentAnalyseResult` , then you can reference them as e.g. `SubscriptionAmendmentAnalyseResult.ReadyToAmend`.  (although in my case it would be `SubscriptionNextStage.ReadyToAmend` )

johnduffell (approved):
nice one, looks good, I did leave a few minor comments which may be useful while I was trying to get my head around it in general

lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
johnduffell: see prototype of this change here - for interest: https://github.com/guardian/price-migration-engine/pull/1515

PR #1508, opened 2026-08-04
Introduce ex_sp2026_notification_active_rateplan_id

shtukas:
In anticipation of the coming introduction of a pause between the Notification step and the Amendment step for SupporterPlus2026, we are introducing a new `CohortItem` attribute called `ex_sp2026_notification_active_rateplan_id`.

We are going to record the active rateplan id at the Notification step and will perform an integrity check at Amendment step. 

This solution is very specific to SupporterPlus2026 which has a higher incidence of people wanting to perform a subscription update after receiving the notification email, but the product switch API does not work if the subscription has been marked with a future price rise amendment. With that said, it can be used more systematically between the Estimation step and the Notification step as a better, more accurate, version of the checks we currently have in place.


lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -267,16 +267,83 @@ sealed trait SubscriptionAmendmentAnalyseResult
...
+  def weHaveSubscriptionRatePlanConsistency(
+      item: CohortItem,
+      subscription: ZuoraSubscription,
+      today: LocalDate
+  ): Option[Boolean] = {
+    // Here we want to check that the subscription main rate plan id has the same value as
+    // attribute `ex_sp2026_notification_active_rateplan_id`
+    for {
+      recordedRatePlanId <-
+        item.ex_sp2026_notification_active_rateplan_id
```
johnduffell, line 292: are you going to backfill these into dynamo to stop it failing "noisily"?

lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -267,16 +267,83 @@ sealed trait SubscriptionAmendmentAnalyseResult
...
+      subscription,
+      today
+    ) match {
+      case None              => Some(SAARFailNoisily)
+      case Some(consistency) =>
+        if (consistency) {
+          Some(SAARReadyToAmend)
+        } else {
+          Some(SAARExcludeFromMigration)
+        }
```
johnduffell, line 316: you might prefer pulling all the conditions up to the match statment e.g.
```suggestion
      case Some(true) =>
          Some(SAARReadyToAmend)
      case Some(false) =>
          Some(SAARExcludeFromMigration)
```

or

```suggestion
      case Some(consistency) if consistency =>
          Some(SAARReadyToAmend)
      case Some(_) =>
          Some(SAARExcludeFromMigration)
```

lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -267,16 +267,83 @@ sealed trait SubscriptionAmendmentAnalyseResult
...
+      item: CohortItem,
+      subscription: ZuoraSubscription,
+      today: LocalDate
+  ): Option[SubscriptionAmendmentAnalyseResult] = {
+    weHaveSubscriptionRatePlanConsistency(
+      item,
+      subscription,
+      today
+    ) match {
+      case None              => Some(SAARFailNoisily)
```
johnduffell, line 310: we won't actually know why it was None (could be missing from dynamo or the sub had no rate plans.)  It's probably easy enough to work out, but it could be captured in an Either or similar.

lambda/src/main/scala/pricemigrationengine/model/AmendmentHandlerHelper.scala
```diff
@@ -267,16 +267,83 @@ sealed trait SubscriptionAmendmentAnalyseResult
...
+        today
+      )
+    } yield recordedRatePlanId == ratePlan.id
+  }
+
+  def analyseSupporterPlus2026(
+      item: CohortItem,
+      subscription: ZuoraSubscription,
+      today: LocalDate
+  ): Option[SubscriptionAmendmentAnalyseResult] = {
```
johnduffell, line 304: this always returns a `Some` so could be simplified

johnduffell (approved):
looks good, just a few thoughts left in the comments, but nothing blocking, mostly style or confusion 👏

PR #1509, opened 2026-08-04
SupporterPlus2026 wait between notification and amendment

shtukas:
This is a follow up of the Introduction of `ex_sp2026_notification_active_rateplan_id` ( https://github.com/guardian/price-migration-engine/pull/1508 ), where we gave ourselves tools to detect subscription divergence between the Notification step and the Amendment step.

Here we introduce the pause between the Notification and the Amendment


johnduffell (approved):
nice! thanks for breaking these into clear separate PRs, it makes it much easier to review (even as someone who has little knowledge of the price rise engine!)

PR #1515, opened 2026-08-05
extract the cohort table update logic

johnduffell:
as per https://github.com/guardian/price-migration-engine/pull/1507#discussion_r3719350058
and this related chat thread: https://chat.google.com/room/AAAAFNJwTZs/klVkH3V68lg/JJXtN3LYqwE?cls=10

This reduces the power of the `processCohortItem` function by stopping it having arbitrary access to the cohort dynamo table.  It has the ability to take exactly one `CohortItem` (as before) but now it can return 0 or 1 `CohortItem` to update back into the table.
(note that the only reason it would return 0 would be if there's a lock contention in zuora, then it will pick it up again on the next attempt)

There are concerns that this would break with the convention of similar functions always returning `Unit` rather than a proper value, but in isolation it makes sense to constrain a function as much as possible by its signature.

### DIFF
--- a/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
+++ b/lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
@@ -152,9 +152,7 @@ object AmendmentHandler extends CohortHandler {
   private def performAmendmentAttempt(
       cohortSpec: CohortSpec,
       item: CohortItem
-  ): ZIO[Zuora with Logging with SalesforceClient, Failure, Option[CohortItem]] = {
-    // This function performs the amendment (through the migration dispatch)
-    // and updates the Cohort Item.
+  ): ZIO[Zuora with Logging with SalesforceClient, Failure, Option[CohortItem]] =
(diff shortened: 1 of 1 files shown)


lambda/src/main/scala/pricemigrationengine/handlers/AmendmentHandler.scala
```diff
@@ -55,6 +55,7 @@ object AmendmentHandler extends CohortHandler {
             Clock.nanoTime.map(_ < deadline)
           )
           .mapZIO(item => processCohortItem(cohortSpec, item))
+          .mapZIO(_.map(CohortTable.update).getOrElse(ZIO.unit))
```
johnduffell, line 58: now the cohort table is only read and written up at this top level, as processCohortItem returns an optional update to the item
So now we can tell that it only updates zero or one item rather than having general table access (ideally it would also enforce that it is only the exact item that it's processing, but there will need to be more hacking of the types to deal with that)

shtukas (approved):
Well Done!

